PR #2, opened 2022-03-22
OSOE-83: Run SCSS pipeline via MSBuild & Publish NodeJs.Extensions to NuGet ⭐

0liver:
OSOE-83


Lombiq.NodeJs.Extensions/package.json
```diff
@@ -3,45 +3,54 @@
...
+    "--": "-----------------------------------------------------------------------------------------------------------",
+    "__shell_docs__": "https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-shell/tests/shell.test.ts"
   },
   "browserslist": [
-    "last 1 Chrome version",
-    "last 1 Edge version",
-    "last 1 Firefox version",
-    "last 1 IE version",
-    "last 1 iOS version"
+    "defaults"
```
0liver, line 57: @Piedone This seems to be a [much better default](https://github.com/browserslist/browserslist#:~:text=defaults%3A%20browserslist%E2%80%99s%20default%20browsers) than what we had before.

Lombiq.NodeJs.Extensions.Samples/package.json
```diff
@@ -1,12 +1,9 @@
 {
   "private": true,
   "devDependencies": {
-    "nodejs-extensions": "link:..\\Lombiq.NodeJs.Extensions",
```
0liver, line 4: @Piedone Removing this as it will be added automatically during the first build.

Lombiq.NodeJs.Extensions.Samples/package.json
```diff
@@ -1,12 +1,9 @@
...
     "npm-run-all": "^4.1.5"
   },
   "scripts": {
-    "dotnet-prebuild": "pnpm build",
-    "dotnet-postclean": "pnpm clean",
```
0liver, line 9: @Piedone This has moved into the main project, now.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
```diff
@@ -0,0 +1,28 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <IsNuGetPackage>$(MSBuildThisFileDirectory.StartsWith($(NuGetPackageRoot)))</IsNuGetPackage>
+    <!-- Referencing 'lombiq.npm.targets' using all lowercase letters, because that's the way its spelled in GitHub
+         Actions which run on Linux, which uses a case-sensitive file system. On Windows this works flawlessly. -->
+    <LombiqNpmTargetsPath
+      Condition="'$(IsNuGetPackage)' == 'true'">$(NuGetPackageRoot.TrimEnd('/\'))/lombiq.npm.targets/1.0.1-alpha.4/build</LombiqNpmTargetsPath>
```
0liver: @Piedone Dropped using `NuGetBuild` here because we need to reference files from _Lombiq.Npm.Targets_ not only during build, but also when installed as a NuGet package.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
```diff
@@ -0,0 +1,33 @@
+<Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+
+  <PropertyGroup>
+    <IsNuGetPackage>$(MSBuildThisFileDirectory.StartsWith($(NuGetPackageRoot)))</IsNuGetPackage>
+    <!-- Referencing 'lombiq.npm.targets' using all lowercase letters, because that's the way its spelled in GitHub
+         Actions which run on Linux, which uses a case-sensitive file system. On Windows this works flawlessly. -->
+    <LombiqNpmTargetsPath
+      Condition="'$(IsNuGetPackage)' == 'true'">$(NuGetPackageRoot.TrimEnd('/\'))/lombiq.npm.targets/1.0.1-alpha.5/build</LombiqNpmTargetsPath>
+    <LombiqNpmTargetsPath
+      Condition="'$(IsNuGetPackage)' != 'true'">$(MSBuildThisFileDirectory.TrimEnd('/\'))/../../lombiq.npm.targets</LombiqNpmTargetsPath>
```
Piedone: Can't this be done otherwise? Even if NPM Targets needs adjusting, or if we need a separate targets file too. This looks like a fragile implementation that we need to take special care of maintaining. We should rather get this working with how props files from NuGet are imported.

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: Are these changes due to this? (It's a .patch file but GH doesn't allow it.) Because during build nothing should change in the consumer project that is or might be stored in its repo and is environment specific (these reference my user folder).

[0001-Package-file-change.txt](https://github.com/Lombiq/NodeJs-Extensions/files/8391738/0001-Package-file-change.txt)

Lombiq.NodeJs.Extensions.Samples.NuGet/Assets/Styles/sample.scss
```diff
@@ -0,0 +1,11 @@
+.banner {
```
Piedone, line 1: Changing this file, or any other SCSS file in the other Samples project or in Privacy, doesn't update the CSS file on build, you need to executed an explicit build from the Task Runner Explorer.

Readme.md
```diff
@@ -2,40 +2,44 @@
...
-
-```
-npm install --save-dev ../../Utilities/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions
-```
-This will be the path if your project resides in a child of the *src* folder. Adjust it to your needs.
-
-Afterwards, you can use any of the [available scripts](#available-scripts) in this package.
+This project can be consumed as a `git` submodule or as a `NuGet` package.
 
+### As a git submodule
```
Piedone:
```suggestion
### As a Git submodule
```

Readme.md
```diff
@@ -2,40 +2,44 @@
...
+In case you've placed the submodule in a different location, adjust the paths as necessary.
+
+### As a NuGet package
+
+When adding _Lombiq.NodeJs.Extensions_ as a NuGet package, no further steps are necessary.
 
-Then, warnings will be sent to the error list if the linter finds rule violations.
+### Integration with MSBuild
+
+_Lombiq.NodeJs.Extensions_ tightly integrates with the MSBuild build system and executes linting and compilation tasks transparently. In case of warnings or errors during the execution of those tasks, respective warnings and errors will be generated with MSBuild and brought to your attention.
```
Piedone: This is a technical identifier (as opposed to "Lombiq Node.js Extensions"), thus needs to be treated as code. Also in the other places.

```suggestion
`Lombiq.NodeJs.Extensions` tightly integrates with the MSBuild build system and executes linting and compilation tasks transparently. In case of warnings or errors during the execution of those tasks, respective warnings and errors will be generated with MSBuild and brought to your attention.
```

Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
```diff
@@ -1,29 +1,22 @@
 <Project Sdk="Microsoft.NET.Sdk.Razor">
 
+  <Import Project="..\Lombiq.NodeJs.Extensions\Lombiq.NodeJs.Extensions.props" />
+
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
     <DefaultItemExcludes>$(DefaultItemExcludes);.git*;node_modules\**</DefaultItemExcludes>
+    <NodeJsExtensionsStylesSourceFolder>NonDefaultScssFolder</NodeJsExtensionsStylesSourceFolder>
+    <NodeJsExtensionsStylesTargetFolder>wwwwNonDefaultCssFolder</NodeJsExtensionsStylesTargetFolder>
```
Piedone, line 9: Add docs about such configs.

Readme.md
```diff
@@ -47,7 +51,9 @@ Here's an overview of all of the scripts this project makes available, categoriz
 
 For even more examples and a mixed setup of default and non-default paths, please check out our dedicated [Samples](Lombiq.NodeJs.Extensions.Samples) project.
 
-To see and run all of the defined scripts in the Visual Studio Task Runner Explorer console, you need to install the [NPM Task Runner](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.NpmTaskRunner64) extension for Visual Studio. You can then run the given scripts and inspect any errors and linter rule violations in the attached console.
+The [NuGet Samples](Lombiq.NodeJs.Extensions.Samples.NuGet) project serves as an example of how to use _Lombiq.NodeJs.Extensions_ from its NuGet package.
```
Piedone: Also for the above Samples link.

```suggestion
The [NuGet Samples](Lombiq.NodeJs.Extensions.Samples.NuGet/Readme.md) project serves as an example of how to use _Lombiq.NodeJs.Extensions_ from its NuGet package.
```

Readme.md
```diff
@@ -61,6 +67,6 @@ If you're using *Node.js* 16.9 or later, you can enable *pnpm* by once executing
 
 ## Contributing and support
 
-Bug reports, feature requests, comments, questions, code contributions, and love letters are warmly welcome, please do so via GitHub issues and pull requests. Please adhere to our [open-source guidelines](https://lombiq.com/open-source-guidelines) while doing so.
+Bug reports, feature requests, comments, questions, code contributions, and love letters are warmly welcome, please submit them via GitHub issues and pull requests. Please adhere to our [open-source guidelines](https://lombiq.com/open-source-guidelines) while doing so.
```
Piedone: This should be the same in every project.

```suggestion
Bug reports, feature requests, comments, questions, code contributions, and love letters are warmly welcome, please do so them via GitHub issues and pull requests. Please adhere to our [open-source guidelines](https://lombiq.com/open-source-guidelines) while doing so.
```

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: Great!

Lombiq.NodeJs.Extensions.Samples/package.json
```diff
@@ -1,20 +1,8 @@
...
-    "build": "npm-run-all --parallel build:styles build:styles:args",
-    "build:styles": "npm explore nodejs-extensions -- pnpm build:styles",
-    "build:styles:args": "npm explore nodejs-extensions -- pnpm build:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder",
-    "clean": "npm-run-all --parallel clean:styles clean:styles:args",
-    "clean:styles": "npm explore nodejs-extensions -- pnpm clean:styles",
-    "clean:styles:args": "npm explore nodejs-extensions -- pnpm clean:styles:args --target=wwwwNonDefaultCssFolder",
-    "watch": "npm-run-all --parallel watch:styles watch:styles:args",
-    "watch:styles": "npm explore nodejs-extensions -- pnpm watch:styles",
-    "watch:styles:args": "npm explore nodejs-extensions -- pnpm watch:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder"
```
Piedone, line 18: Why not keep these here for demonstration?

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -4,14 +4,46 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <DefaultItemExcludes>$(DefaultItemExcludes);.*;node_modules\**</DefaultItemExcludes>
     <IsPublishable>false</IsPublishable>
-    <InitNodeModulesStampFile>node_modules\.install-stamp</InitNodeModulesStampFile>
+    <!-- The NuGet package reference to Lombiq.Npm.Targets would run those targets during the build of this project,
```
Piedone: Isn't the source build running them too?

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -4,14 +4,46 @@
...
+    <ExecNpmInstallCommand>false</ExecNpmInstallCommand>
+    <ExecDotnetPostcleanCommand>false</ExecDotnetPostcleanCommand>
+    <ExecDotnetPrebuildCommand>false</ExecDotnetPrebuildCommand>
+  </PropertyGroup>
+
+  <PropertyGroup>
+    <Title>Lombiq Node.js Extensions</Title>
+    <Authors>Lombiq Technologies</Authors>
+    <Copyright>Copyright © 2022, Lombiq Technologies Ltd.</Copyright>
+    <Description> See the project website for detailed documentation.</Description>
```
Piedone: Description missing.

Lombiq.NodeJs.Extensions/package.json
Piedone: Yep.

Piedone (changes requested):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
0liver: Well, `.props` files from NuGet packages are only imported into projects from immediate references, not transitive ones. Without pulling in the `lombiq.npm.targets` here, they won't be imported at all.

What exactly is it that you deem fragile?

Lombiq.NodeJs.Extensions.Samples.NuGet/Assets/Styles/sample.scss
0liver: I'll have to investigate that. I've tried many things and don't recall this problem.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
Piedone: `IncludeAssets` and `PrivateAssets` affect this. Can't they be used?

Instead of relying on NuGet to import everything we need, we're trying to look up the file in the package cache. This is working around the features NuGet provides for this scenario, and that shouldn't be necessary. Such things are alway risky, and here we also need to keep the version number up to date, something that we'll forgot a lot (since updating NuGet dependencies doesn't necessarily involve looking at the csproj file).

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Well, we have all the scripts this project needs below. The scripts without `:args` are referenced in the `Samples.NuGet` project.

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Yes.

I'm coming to think the installation of NE as an `npm package` will have to be something that consumers will have to do by hand, using an MSBuild target (because we may need to access the `NuGetPackageRoot`). Then, the changes to _package.json_ and `pnpm-lock.yaml` will be self-produced and obviously necessary to commit.

With that we would avoid adjusting those files during build.

Hm, I'm not yet sure how to reference the extension from the NuGet cache in a way that will be user-agnostic 🤔

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
0liver: I've never heard of `IncludeAssets` and `PrivateAssets,` so I'll go and change that 😁

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Maybe the `PrivateAssets` will help me out here.

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: But why not also have `build:styles` and the likes?

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: That it adds files to the consumer project is fine, and if the consumer needs to be changed in any case, then doing this automatically is great. However, it can't include environment-specific things. If this is the case even during a manual installation, then we need to figure out something.

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Because those use the default locations for SCSS and CSS files. This sample project only uses non-default paths. As discussed in JIRA, we don't need support for both at once.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: No, because we don't include the `Lombiq.Npm.Targets` build files in `Lombiq.NodeJs.Extensions.` But with the NuGet package reference the `Lombiq.Npm.Targets` build files are automagically loaded for `Lombiq.NodeJs.Extensions.` and would execute their targets.

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: In case of the submodule usage, I can try to use a relative path. For the NuGet package usage, I still have to investigate.

Lombiq.NodeJs.Extensions.Samples.NuGet/Assets/Styles/sample.scss
0liver: I've found the problem and fixed it 😀

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: We can define `build:styles` here to work still though. My issue is, that having such separate build and watch tasks is a common use case, so we need to also demonstrate them.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: Why are we not including them when doing the source build though? In effect, NE should work the same from source and NuGet too. If we need NPM Targets for that, then I don't understand why we'd rely on the NPM Targets' targets file to be active during NuGet build but not during source build.

0liver: NE does not need NPM Targets itself. NE needs NPM Targets' `.props` and `.targets` files to build consumer projects.

Now, the NuGet package reference to NPM Targets automatically includes NPM Targets' `.props` and `.targets` files when NE is built, which we are working around here. I don't know of a better way to do that. `<ExcludeAssets>build</ExcludeAssets>` would probably achieve that, but we need the build assets (`.props` and `.targets` files) for the consumer projects.

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: I'm arguing that having both `build:styles` AND `build:styles:args` scripts is NOT a common case, because it would imply that the consumer project has both the default AND a non-default location with SCSS files.

As discussed, we actually don't provide MSBuild integration for BOTH at the same time. You either use the default locations _Assets/Styles_ and _wwwroot/css_ or you use something else. In the former case, you'd have:
```js
"build:styles": "npm explore nodejs-extensions -- pnpm build:styles"
```
and in the latter:
```js
"build:styles:args": "npm explore nodejs-extensions -- pnpm build:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder"
```
It's either, or.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: I see. Then elaborate this in the comments something along the lines of, how I understand, "during the build of this project in the beginning (?) automatically. However, we want to run them at specific stages. Thus, we need to disable the automatic execution here."

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: I see. How about having just `build:styles`, `clean:styles` and `watch:styles` though?

0liver: That sounds sensible. I'll do that.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: It's not really about the beginning or other build steps. It's about **this project** being build vs. the **consumer project** being built, and we want the execution of the NPM Targets only in the latter case, **not** the former.

But maybe this will become obsolete once I've figured out the right combination of `PrivateAssets` and `IncludeAssets`.

Piedone: In any case, if this remains, please document the reasons for having it.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
0liver: OK, so I've done some investigation and tests and I am a bit frustrated at this point because:
1. [This paragraph](https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#:~:text=Note%20that%20because,that%20consume%20AppLogger.) reads to me like I should be able to provide `NPM Targets` _.props_ and _.targets_ files to consumer projects by specifying the following in `Lombiq.NodeJs.Extensions.csproj`: 
```xml
    <PackageReference Include="Lombiq.Npm.Targets" Version="1.0.1-alpha.6">
      <!-- We want to make NPM-Targets' .props and .targets files available to consumer projects. -->
      <IncludeAssets>all</IncludeAssets>
      <!-- We don't want to consume neither build not compile assets in NE itself. -->
      <ExcludeAssets>build;compile</ExcludeAssets>
      <!-- We don't want the NPM-Targets DLL in the consumer projects. -->
      <PrivateAssets>compile</PrivateAssets>
    </PackageReference>
```
2. But it simply does not work. I've tried a myriad of values, rebuilt the project, repacked the NuGet package (locally), installed the latest package - the NPM Targets build files are not included in the generated `.nuget.g.props` and `.nuget.g.targtes` files.

Any ideas?

Piedone: I suggest posting a question here: https://github.com/dotnet/msbuild/discussions or on Stackoverflow.

0liver: OK, I've got the inclusion of .props and .targets to work - after adjusting the `.nuspec` file in the locally created package 😁 There seem to be a few bugs during its creation 🤔

0liver: I've filed [an issue](https://github.com/NuGet/Home/issues/11717) now that I've gathered more info on this. It seems quite impactful!

Piedone: Great, let's see the feedback. Can we override just this from a custom nuspec file as a workaround?

0liver: I had the same idea, but since I've never worked with them, I don't know if it's going to work. I will try it out today.

0liver: Looks like the _.nuspec_ file cannot override anything - it has to be used **in place of** the project file. See [the docs](https://docs.microsoft.com/en-us/nuget/reference/nuspec#project-type-compatibility).

Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
0liver: Would you put that documentation into the main _Readme.md_ or into _Docs/Styles.md_? It is specific for styles, but I'm wondering if it's not important enough to be in the main readme.

Piedone: Into Styles.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.props
0liver: After trying for about 2h and ever new errors popping up, I gave up on tuning the `.nuspec` file by hand.

I have a working configuration now which include the `Npm.Targets` build files before those from `NodeJs.Extensions` 🎉

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: I've been able to remove this overrides now 🎉

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: We can keep the relative path here. When using the NE NuGet package, though, its location is always kind of environment-specific, as it uses the system's NuGet packages folder. What now?

Piedone: I don't have an existing answer for this. This is one of the fundamental questions why I've opened https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/48 in the first place.

0liver: I'm thinking that we could copy the NE _package.json_ into a subfolder of the consuming project, and `npm install` it from there - then all of the work will happen inside of that folder. That would make us independent of the NuGet folder structure.

Piedone: I'm OK with that.

0liver: I've got that working now.

Piedone: Great!

0liver: The automatic reference now points to `node_modules\.nx`. I've chosen the short `.nx` segment because I've hit a problem when some dependent npm package's installation path hit the 260 character boundary. I now that you can enable support for longer paths on Windows, and I do have that enabled. This part will still fail in that case.

I've added according documentation under the **Troubleshooting** section in the _Readme_.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: I've had to restore the overrides, and I've documented why. It's a separate block at the bottom of the file now.

Lombiq.NodeJs.Extensions.Samples/.stylelintrc
```diff
@@ -0,0 +1,6 @@
+{
```
Piedone, line 1: There are created for the NuGet project too:

image

(The pnpm-lock in the base project shows a change without actually changing; I suspect it has something to do with EOL - https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/issues/65 - but Git doesn't show a change.)

Readme.md
```diff
@@ -13,34 +13,48 @@ This project allows you to use predefined build scripts for *SCSS* files without
...
+
+### Troubleshooting
 
-During the first build of your project after adding _Lombiq.NodeJs.Extensions_, it will additionally be added as an `npm` package to your project, which allows you to run the contained `npm` scripts from your project. Refer to the [available scripts](#available-scripts) section for more information.
+Should you encounter the following error:
+
+```
+ΓÇëERRORΓÇë ENOENT: no such file or directory, realpath [...]
+```
+... then please try and override the `NodeJsExtensionsNpmPackageSourcePath` property with something shorter than the default value of `./node_modules/.nx`. You can try to use `.nx`, but you then need to add *.nx* to your *.gitignore* file. Should that still not work, please use the relative path to your solution root or something similar.
```
Piedone:
```suggestion
...then please try and override the `NodeJsExtensionsNpmPackageSourcePath` property with something shorter than the default value of `./node_modules/.nx`. You can try to use `.nx`, but you then need to add *.nx* to your *.gitignore* file. Should that still not work, please use the relative path to your solution root or something similar.
```

Readme.md
Piedone: You can just commit these suggestions so they're auto-resolved.

Lombiq.NodeJs.Extensions.Samples/.stylelintrc
0liver: Should I just check these in? I didn't do that on purpose, because I thought they would be part of the learning effect.

I don't have that change of the _pnpm-lock_ file in the `Samples` project.

Readme.md
0liver: Can you point to a source for this suggestion?

Piedone: Weird that you see this already because I didn't finish and send my review yet.

There seems to be no definite spelling rule on spaces before/after (or even within) an ellipsis, this being up to the given style. However, some indications point to it not being valid as the start of a sentence anyway, so I'll add a different suggestion.

Lombiq.NodeJs.Extensions.Samples/.stylelintrc
Piedone: Yes, invoking a build shouldn't cause the clone to become dirty.

Readme.md
```diff
@@ -13,34 +13,48 @@ This project allows you to use predefined build scripts for *SCSS* files without
...
 
+The underlying problem is a too long path name, and the error appears even when the support for path lengths of over 260 characters has been enabled.
```
Piedone:
```suggestion
You may encounter the following error:

\```
ΓÇëERRORΓÇë ENOENT: no such file or directory, realpath [...]
\```

In this case, please try moving your solution to a folder with a shorter path. Should this not be enough, try to override the `NodeJsExtensionsNpmPackageSourcePath` property with something shorter than the default value of `./node_modules/.nx`. You can try to use `.nx`, but you then need to add *.nx* to your *.gitignore* file. If this still doesn't work, please use the relative path to your solution root or something similar.

The underlying problem is a too long path name, and the error appears even when the support for path lengths of over 260 characters has been enabled.
```

Readme.md
```diff
@@ -13,34 +13,48 @@ This project allows you to use predefined build scripts for *SCSS* files without
...
+
+During the first build of your project after adding `Lombiq.NodeJs.Extensions`, it will additionally be added as an `npm` package to your project, which allows you to run the contained `npm` scripts from your project. Refer to the [available scripts](#available-scripts) section for more information.
+
+### Troubleshooting
 
-During the first build of your project after adding _Lombiq.NodeJs.Extensions_, it will additionally be added as an `npm` package to your project, which allows you to run the contained `npm` scripts from your project. Refer to the [available scripts](#available-scripts) section for more information.
+Should you encounter the following error:
+
+```
+ΓÇëERRORΓÇë ENOENT: no such file or directory, realpath [...]
```
Piedone, line 53: Add a note about this to https://lombiq.atlassian.net/wiki/spaces/DEV/pages/1748369460/Setting+up+Node.js+with+Visual+Studio as well.

Lombiq.NodeJs.Extensions.Samples/package.json
```diff
@@ -1,8 +1,11 @@
 {
   "private": true,
   "scripts": {
-    "build": "npm explore nodejs-extensions -- pnpm build:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder",
-    "clean": "npm explore nodejs-extensions -- pnpm clean:styles:args --target=wwwwNonDefaultCssFolder",
-    "watch": "npm explore nodejs-extensions -- pnpm watch:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder"
+    "build:styles": "npm explore nodejs-extensions -- pnpm build:styles:args --source=NonDefaultScssFolder --target=wwwwNonDefaultCssFolder",
```
Piedone, line 4: This doesn't get to run when changing an SCSS file and hitting F6 (given the CSS file already exists).

Docs/Styles.md
```diff
@@ -54,8 +64,10 @@ The rules are found in 2 files:
 - *lombiq-base.stylelintrc.json*: These rules are Lombiq overrides for [stylelint-config-standard-scss](https://www.npmjs.com/package/stylelint-config-standard-scss).
 - *.stylelintrc*: In this file you can define your own overriding rules.
 
+The *.stylelintrc* file will automatically be created in your project during the first build, and includes *lombiq-base.stylelintrc.json* from the Node.js Extensions `npm` package. Should you be using, or want to use, a global *.stylelintrc* for your whole solution, or use any other way of [configuring Stylelint](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/configure.md#configuration), you can disable this behavior by setting `<NodeJsExtensionsCreateStylelintConfigurationFile>` to `false` in your project file.
```
Piedone:
```suggestion
The *.stylelintrc* file will automatically be created in your project during the first build, and extends *lombiq-base.stylelintrc.json* from the Node.js Extensions `npm` package. Should you be using, or want to use, a global *.stylelintrc* for your whole solution, or use any other way of [configuring Stylelint](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/configure.md#configuration), you can disable this behavior by setting `<NodeJsExtensionsCreateStylelintConfigurationFile>` to `false` in your project file.
```

Piedone: How would one use a global stylelintrc in a solution? Document it here.

Docs/Styles.md
```diff
@@ -2,39 +2,49 @@
 
 
 
-These scripts lint (with [Stylelint](https://stylelint.io/)), compile, autoprefix, and minify the given *scss* files into an output folder. Beyond that, there are also a *clean*  and a *watch* script.
```
Piedone, line 5: Add this file under the Lombiq.NodeJs.Extensions folder instead of the root, so we can manage it from inside the project and VS:

Docs/Styles.md
```diff
@@ -2,39 +2,49 @@
...
-3. For even more examples and a mixed setup of default and non-default paths, please check out our dedicated [Samples](../Lombiq.NodeJs.Extensions.Samples) project.
-
-To see and run all of the defined scripts in the Visual Studio Task Runner Explorer console, you need to install the [NPM Task Runner](https://marketplace.visualstudio.com/items?itemName=MadsKristensen.NpmTaskRunner64) extension for Visual Studio. You can then run the given scripts and inspect any errors and linter rule violations in the attached console.
+
+## Source and target paths
+
+The default paths for SCSS input files and CSS output files are *Assets/Styles* and *wwwroot/css*, respectively. The existing folder structure in the input folder will be mirrored in the output, e.g. *Assets/Styles/app/main.scss* will be transformed into *wwwroot/css/app/main.css*, together with *wwwroot/css/app/main.min.css* and *wwwroot/css/app/main.css.map*.
+
+Those defaults can be overridden by providing the following MSBuild properties in your project file:
+```xml
```
Piedone: Surround fenced code blocks with line breaks.

Docs/Styles.md
```diff
@@ -2,39 +2,49 @@
...
+The below scripts lint (with [Stylelint](https://stylelint.io/)), compile, autoprefix, and minify the given SCSS files into an output folder. Beyond that, there are also `clean` and `watch` scripts.
+
+The usage of these scripts is optional. `Lombiq Node.js Extensions` is able to process your SCSS files during the regular build of your project via MSBuild or the `dotnet` CLI without further work from your side. These scripts are meant to be used during development for short feedback loops, most of all the `watch` script.
 
 Looking for something similar for .NET? Check out our [.NET Analyzers project](https://github.com/Lombiq/.NET-Analyzers).
 
-You can use it as follows:
-
-1. Copy *example.stylelintrc* from the *Stylelint* folder of this project to the root folder of your solution (i.e. where you have the *.sln* file), rename it to *.stylelintrc*, and optionally adjust the location of *lombiq-base.stylelintrc* inside the file.
-2. In your *package.json*, add any or all of the following entries to the `scripts` property:
```
Piedone, line 12: Even if this is not necessary anymore, add a step-by-step guide on how to start with utilizing this in your own project.

Readme.md
```diff
@@ -13,34 +13,48 @@ This project allows you to use predefined build scripts for *SCSS* files without
...
 
 
 ## Installation and usage
 
 This project can be consumed as a `git` submodule or as a `NuGet` package.
 
-### As a git submodule
-In the case of using it as a git submodule, it is recommended to put the [main](../Lombiq.NodeJs.Extensions) project into a folder named _Lombiq.NodeJs.Extensions_ under the _src/Utilities_ folder, but this is not mandatory.
+### As a Git submodule
+In the case of using `Lombiq.NodeJs.Extensions` as a git submodule, it is recommended to put it into a folder named _Lombiq.NodeJs.Extensions_ under the _src/Utilities_ folder, but you are free to use a differnt location. You need to add [`Lombiq.Npm.Targets`](https://github.com/Lombiq/NPM-Targets) to the same folder, though.
```
Piedone:
```suggestion
In the case of using `Lombiq.NodeJs.Extensions` as a git submodule, it is recommended to put it into a folder named _Lombiq.NodeJs.Extensions_ under the _src/Utilities_ folder, but you are free to use a different location. You need to add [`Lombiq.Npm.Targets`](https://github.com/Lombiq/NPM-Targets) to the same folder, though.
```

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -42,8 +37,29 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(NuGetBuild)' == 'true'">
-    <!-- The version used here needs to be mirrored in the <LombiqNpmTargetsPath> property in the .props file. -->
-    <PackageReference Include="Lombiq.Npm.Targets" Version="1.0.1-alpha.5" />
+    <PackageReference Include="Lombiq.Npm.Targets" Version="1.0.1-alpha.14">
+      <!-- See https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#controlling-dependency-assets.
+           and https://github.com/NuGet/docs.microsoft.com-nuget/issues/2716. -->
+      <!-- We want to make NPM-Targets' build files (.props and .targets) available to consumer projects, so we need to
+           specify "build" here. We also need "buildTransitive", because otherwise it will become part of the "exlude"
```
Piedone:
```suggestion
           specify "build" here. We also need "buildTransitive", because otherwise it will become part of the "exclude"
```

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -42,8 +37,29 @@
...
   </ItemGroup>
 
+  <PropertyGroup>
+    <!-- Suppress execution of NPM Targets during the build of this project. It would be nicer to solve this with
+         ExcludeAssets on the "Lombiq.Npm.Targets" PackageReference, but that does not work. There is currently no way
+         for PackageReferences to specify that some assets should not be consumed by itself but still forwarded to
+         projects consuming this one. -->
+    <ExecNpmInstallCommand>false</ExecNpmInstallCommand>
+    <ExecDotnetPostcleanCommand>false</ExecDotnetPostcleanCommand>
+    <ExecDotnetPrebuildCommand>false</ExecDotnetPrebuildCommand>
```
Piedone, line 62: Don't these overrides take affect for all consumer projects too, at least if used in source form?

Piedone (changes requested):

Readme.md
0liver: Looks like the backticks have to be escaped:

image

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: What is `F6` the shortcut for?

Also, these scripts are only run via either the Task Runner, or from a shell via `npm run build:styles` etc.

Piedone: Build. We need to run SCSS build during .NET build, this is the whole point of NPM Targets (and something you confirmed working [here](https://github.com/Lombiq/NodeJs-Extensions/pull/2/files#diff-a442051efc5091082fa3ee450aa4617a17730c261db904844cefec329c380198)).

Readme.md
Piedone: Fixed it but note that the backticks need to be removed in the code block (can't see to do otherwise).

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: No, because the project file is not included in consumer projects, so these properties won't be set when building the consumer project.

Docs/Styles.md
0liver: I've added a headline "How to get started" above the paragraph with the instructions I've move down a bit.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: Doesn't a project reference cause that though?

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Of course!

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: No. A project reference only dictates the build order and the use of assets.

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: Hm. The following works for me:

1. Build solution.
2. Change SCSS file in `Lombiq.Privacy`.
3. Build solution again (not Rebuild).
4. The CSS file is correctly updated.

Are you saying that this does not work for you? If yes, please change the verbosity for the build to detailed and send me the log. You can save that by hitting `CTRL` + `S` after focusing the Output window.

Readme.md
0liver: [Here it is.](https://lombiq.atlassian.net/wiki/spaces/DEV/pages/1748369460/Setting+up+Node.js+with+Visual+Studio#%CE%93%C3%87%C3%ABERROR%CE%93%C3%87%C3%AB-ENOENT%3A-no-such-file-or-directory%2C-realpath-%5B...%5D)

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: Great then.

Lombiq.NodeJs.Extensions.Samples/package.json
Piedone: This works in Privacy, but not in the Samples (I commented on the Samples) with nested-samples (now I noticed that it does work with non-default-samples). Do check how this is in the NuGet sample too.

This is the output after changing nested-samples:

[Output-Build.zip](https://github.com/Lombiq/NodeJs-Extensions/files/8521789/Output-Build.zip)

Docs/Styles.md
```diff
@@ -14,13 +14,19 @@ Looking for something similar for .NET? Check out our [.NET Analyzers project](h
 The default paths for SCSS input files and CSS output files are *Assets/Styles* and *wwwroot/css*, respectively. The existing folder structure in the input folder will be mirrored in the output, e.g. *Assets/Styles/app/main.scss* will be transformed into *wwwroot/css/app/main.css*, together with *wwwroot/css/app/main.min.css* and *wwwroot/css/app/main.css.map*.
 
 Those defaults can be overridden by providing the following MSBuild properties in your project file:
+
```
Piedone: There are now two Styles.md files.

Lombiq.NodeJs.Extensions/Docs/Styles.md
```diff
@@ -0,0 +1,91 @@
...
+The *.stylelintrc* file will automatically be created in your project during the first build, and extends *lombiq-base.stylelintrc.json* from the Node.js Extensions `npm` package. Should you prefer to use a global *.stylelintrc* for your whole solution, or use any other way of [configuring Stylelint](https://github.com/stylelint/stylelint/blob/main/docs/user-guide/configure.md#configuration), you can disable this behavior by adding the following property to your project file:
+
+```xml
+<NodeJsExtensionsCreateStylelintConfigurationFile>false</NodeJsExtensionsCreateStylelintConfigurationFile>
+```
+
+You can use a single *.stylelintrc* configuration file for all projects in a solution as follows:
+
+1. Move *.stylelintrc* from your project into the root folder of your solution, i.e. next to your solution file.
+2. Edit *.stylelintrc* and adjust the path to *lombiq-base.stylelintrc.json*.
```
Piedone, line 85: We should also do this.

Piedone (changes requested):

Docs/Styles.md
0liver: Now, not anymore!

Lombiq.NodeJs.Extensions.Samples/package.json
0liver: This is now fixed 😀

Piedone: Great!

Lombiq.NodeJs.Extensions/package.json
```diff
@@ -15,8 +15,9 @@
     "stylelint-config-standard-scss": "^3.0.0"
   },
   "scripts": {
-    "dotnet-prebuild": "pnpm build:styles",
-    "dotnet-postclean": "pnpm clean:styles",
+    "build": "pnpm build:styles",
+    "compile": "pnpm compile:styles",
+    "clean": "pnpm clean:styles",
```
Piedone, line 20: Why where these changes necessary?

0liver: Since in the `.props` file I call either `build` or `compile`, depending on the value of `$(NuGetBuild)`, it seemed a good idea to use these short script names instead of `dotnet-prebuild` and `dotnet-postclean`. Choosing `build` over `build:styles` also keeps it open to extension by JS compilation in the future.

Piedone (approved):

PR #3, opened 2022-05-31
OSOE-125: Add JavaScript processing functionality

0liver:
OSOE-125


.gitattributes
```diff
@@ -0,0 +1 @@
+* text=auto
```
Piedone, line 1: Use the latest of this: https://github.com/Lombiq/Open-source-template/blob/dev/.gitattributes

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
+# Scripts for JS files
+
+
+
+The below scripts lint (with [ESLint](https://eslint.org/)), transmpile to ES5, and minify the given JS files into an output folder. Beyond that, there are also `clean` and `watch` scripts.
+
+The usage of these scripts is optional. `Lombiq Node.js Extensions` is able to process your SCSS files during the regular build of your project via MSBuild or the `dotnet` CLI without further work from your side. These scripts are meant to be used during development for short feedback loops, most of all the `watch` script.
+
+Looking for something similar for .NET? Check out our [.NET Analyzers project](https://github.com/Lombiq/.NET-Analyzers).
```
Piedone:
```suggestion
The scripts below lint (with [ESLint](https://eslint.org/)), transpile to ES5, and minify the given JS files into an output folder. Beyond that, there are also `clean` and `watch` scripts.

The usage of these scripts is optional. `Lombiq Node.js Extensions` is able to process your JS files during the regular build of your project via MSBuild or the `dotnet` CLI without further work from your side. These scripts are meant to be used during development for short feedback loops, most of all the `watch` script.

Looking for something similar for .NET? Check out our [.NET Analyzers project](https://github.com/Lombiq/.NET-Analyzers).
```

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
...
+To use ESLint in Visual Studio Code, you can use e.g. Microsoft's official [ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).
+
+
+## ESLint rules
+
+The rules are found in 2 files:
+- *eslint.lombiq-base.js*: These rules are Lombiq overrides for [airbnb-base](https://www.npmjs.com/package/eslint-config-airbnb-base).
+- *eslint.config.js*: In this file you can define your own overriding rules.
+
+### Customizing ESLint rules - coming soon!
```
Piedone: Please add this too. This should cover what's under https://github.com/Lombiq/Gulp-Extensions/blob/dev/Docs/JsTargets.md#eslint-rules.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
...
+The usage of these scripts is optional. `Lombiq Node.js Extensions` is able to process your SCSS files during the regular build of your project via MSBuild or the `dotnet` CLI without further work from your side. These scripts are meant to be used during development for short feedback loops, most of all the `watch` script.
+
+Looking for something similar for .NET? Check out our [.NET Analyzers project](https://github.com/Lombiq/.NET-Analyzers).
+
+
+## Source and target paths
+
+The default paths for JS input and output files are *Assets/Scripts* and *wwwroot/js*, respectively. The existing folder structure in the input folder will be mirrored in the output, e.g. *Assets/Scripts/app/main.js* will be transformed into *wwwroot/js/app/main.js*, together with *wwwroot/js/app/main.min.js* and *wwwroot/js/app/main.js.map*.
+
+### Overriding the defaults - coming soon!
```
Piedone: Please add this too.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
...
+    "watch": "npm run watch:scripts",
+    "watch:scripts": "npm explore nodejs-extensions -- npm run watch:scripts:args --source=path/to/my/js-files --target=path/to/my/js-files",
+  }
+  ```
+
+To see the different configurations of default and non-default paths in action, please check out our dedicated [Samples](../../Lombiq.NodeJs.Extensions.Samples/Readme.md) [projects](../../Lombiq.NodeJs.Extensions.Samples.NuGet/Readme.md).
+
+### Integration with Visual Studio (Code)
+
+Visual Studio support ESLint out of the box. You can enable it by ticking the checkbox "Enable ESLint" under *Tools -> Options -> Text Editor -> JavaScript/TypeScript -> Linting -> General*.
```
Piedone: This doesn't fully replicate https://github.com/Lombiq/Gulp-Extensions/blob/dev/Docs/JsTargets.md#:~:text=If%20you%20use,the%20lines%20of%3A

Lombiq.NodeJs.Extensions/pnpm-lock.yaml
```diff
@@ -1,8 +1,17 @@
 lockfileVersion: 5.3
 
 specifiers:
+  '@babel/cli': ^7.17.10
```
Piedone: After a build, the pnpm-lock files of all three projects (this one, UI Kit, Privacy) show up as changed in Git. There's no actually visible change, and when stating them they vanish:

https://user-images.githubusercontent.com/1976647/171283072-9570747a-75d4-4120-b8c1-45f66ea74122.mp4

This is the same as with SCSS files built into wwwroot folders kept in the repos. I suspect it's a line ending issue despite it not showing up as a diff. Perhaps nomalizing line Git line endings will help with this, or removing and reading the files.

Note that this only happens after the first build of a cleaned repo.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
...
+- *eslint.config.js*: In this file you can define your own overriding rules.
+
+### Customizing ESLint rules - coming soon!
+
+
+## Operating System Compatibility Regarding Git and Line Breaks
+
+For historical reasons, Windows uses the `\r\n` character combination (also known as CR-LF) to denote a line break, while Unix-like operating systems such as Linux and macOS simply use a single `\n` character (LF). Git (made by the creator of Linux) treats the Unix-style line endings as the only right option. If you are on Windows your Git client is almost certainly configured to "Checkout Windows-style, commit Unix-style" by default to overcome this cultural difference, but if not then it's a good practice to [configure Git](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace) to ensure your line endings are consistent. We've disabled the [`linebreak-style`](https://eslint.org/docs/rules/linebreak-style) rule to avoid cross compatibility issues.
+
+To ensure that the files have consistent line endings in the remote repository, you can add the following _.gitattributes_ file:
```
Piedone:
```suggestion
To see the different configurations of default and non-default paths in action, please check out our dedicated [Samples](../../Lombiq.NodeJs.Extensions.Samples/Readme.md) [projects](../../Lombiq.NodeJs.Extensions.Samples.NuGet/Readme.md).

### Integration with Visual Studio (Code)

Visual Studio supports ESLint out of the box. You can enable it by ticking the checkbox "Enable ESLint" under *Tools → Options → Text Editor → JavaScript/TypeScript → Linting → General*.

To use ESLint in Visual Studio Code, you can use e.g. Microsoft's official [ESLint plugin](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint).


## ESLint rules

The rules are found in 2 files:
- *eslint.lombiq-base.js*: These rules are Lombiq overrides for [airbnb-base](https://www.npmjs.com/package/eslint-config-airbnb-base).
- *eslint.config.js*: In this file you can define your own overriding rules.

### Customizing ESLint rules - coming soon!


## Operating System Compatibility Regarding Git and Line Breaks

For historical reasons, Windows uses the `\r\n` character combination (also known as CR-LF) to denote a line break, while Unix-like operating systems such as Linux and macOS simply use a single `\n` character (LF). Git (made by the creator of Linux) treats the Unix-style line endings as the only right option. If you are on Windows your Git client is almost certainly configured to "Checkout Windows-style, commit Unix-style" by default to overcome this cultural difference, but if not then it's a good practice to [configure Git](https://git-scm.com/book/en/v2/Customizing-Git-Git-Configuration#_formatting_and_whitespace) to ensure your line endings are consistent. We've disabled the [`linebreak-style`](https://eslint.org/docs/rules/linebreak-style) rule to avoid cross compatibility issues.

To ensure that the files have consistent line endings in the remote repository, you can add the following _.gitattributes_ file:
```

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -0,0 +1,84 @@
+# Scripts for JS files
```
Piedone, line 1: Extend the normal and NuGet sample with JS processing.

Lombiq.NodeJs.Extensions/package.json
```diff
@@ -3,55 +3,78 @@
...
-    "_error:styles": "echo \"Error in styles pipeline: $ERRORSTEP failed.\"",
-    "--": "-----------------------------------------------------------------------------------------------------------",
-    "__shell_docs__": "https://github.com/yarnpkg/berry/blob/master/packages/yarnpkg-shell/tests/shell.test.ts"
+    "---": "------------------------------------------------- SCRIPTS ------------------------------------------------",
+    "build:scripts": "run-p --print-label --continue-on-error lint:scripts minify:scripts",
+    "lint:scripts":
+      "(TARGET_DIR=$(node scripts/get-target) && node scripts/exists $TARGET_DIR/Assets/Scripts || exit 0 && cd $TARGET_DIR npx eslint Assets/Scripts --max-warnings 0 --config ${INIT_CWD}/config/eslint.config.json) || AREA=scripts STEP=linting pnpm _error",
+    "minify:scripts": "run-s --print-label compile:scripts _minify:scripts",
+    "compile:scripts":
+      "(TARGET_DIR=$(node scripts/get-target) && node scripts/exists $TARGET_DIR/Assets/Scripts || exit 0 && cd $TARGET_DIR && babel Assets/Scripts --out-dir wwwroot/js --config-file ${INIT_CWD}/config/babel.config.json) || AREA=scripts STEP=compiling pnpm _error ",
```
Piedone: These lines are really long, break them up a bit.

Lombiq.NodeJs.Extensions/scripts/minify.js
```diff
@@ -0,0 +1,43 @@
...
+
+console.debug(`Minifying ${process.cwd()}...`);
+
+fs.readdir('.', (error, files) => {
+    files.forEach(file => {
+        if (!file.endsWith('.js') || file.endsWith('.min.js')) return;
+
+        const destination = path.basename(file, '.js') + '.min.js';
+
+        // Prefer the terser CLI for simplity over the API. See https://github.com/terser/terser for more information.
```
Piedone:
```suggestion
        // Prefer the terser CLI for simplicity over the API. See https://github.com/terser/terser for more information.
```

Piedone (changes requested):

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: The remainder of the linked section is obsolete now.

Piedone: Previously, VS needed a separate file in a specific location for JS linting to work. Isn't that the case today?

0liver: I've got it to work at one point, but after a full clean, it's broken again. So, yes, it's possible without extra files, I just have to make it permanent.

Piedone: OK, let's see it then.

Lombiq.NodeJs.Extensions/package.json
0liver: I've shortened them, but there's no way to break them up. It'd be possible to defer parts of the call chain to another script entry, but every invocation comes with a not-so-small cost, so I'm trying to avoid unnecessary calls.

Piedone: Why is there no way to break them?

0liver: Because in JSON, keys and values must be string constants.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: Should I do anything with the compiled, minified JS files, like use them on a page? Or can I spare that part?

Piedone: No need for that.

Lombiq.NodeJs.Extensions/package.json
Piedone: I see, of course.

0liver: In this longest line of all (showing the current version) I'm still considering two things:
```json
"compile:scripts":
  "SRC=$(node scripts/get-path js source) DEST=$(node scripts/get-path js target) && babel $SRC --out-dir $DEST --config-file ${INIT_CWD}/config/babel.config.json || AREA=scripts STEP=compilation pnpm error",
```

1. Renaming the `scripts` folder to `_` which would be shorter, somehow OK due to `underscore.js` using the same, and those scripts being internal, anyways. That would save **12 chars** here.
```json
"compile:scripts":
   "SRC=$(node _/get-path js source) DEST=$(node _/get-path js target) && babel $SRC --out-dir $DEST --config-file ${INIT_CWD}/config/babel.config.json || AREA=scripts STEP=compilation pnpm error",
```
2. Moving _babel.config.json_ into the parent folder, which would be `Lombiq.NodeJs.Extensions`' root folder. Then, I wouldn't have to pass the path to the config explicitly.
```json
"compile:scripts":
   "SRC=$(node scripts/get-path js source) DEST=$(node scripts/get-path js target) && babel $SRC --out-dir $DEST || AREA=scripts STEP=compilation pnpm error",
``` 
3. Moving _babel.config.json_  into the consumer project, just like _.stylelintrc_ and _.eslintrc_. This would have the same effect as 2. with the addition that it would allow to easily opt out of transpilation to ES2015 by removing parts of its content. I've wanted to do this initially, but then deemed it out of scope.

What do you think?

Piedone: I wouldn't do 1 since it hurts understandability. 3 sounds good instead of 2.

0liver: Great! I agree 😀

Lombiq.NodeJs.Extensions/pnpm-lock.yaml
0liver: I've seen this too, and have been unable to pinpoint the root cause 🤔

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: Sorry, I didn't commit your suggestion here. I had changed the text in the meantime, so the suggestion was out of date before I saw it.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: The solution-wide `.eslintrc` might be a nice feature, but it has doubled my amount of grey hair during the past days. I've now come to the conclusion that non-project level configuration needs the upfront build of GE, which I **really** wanted to ditch with NE. I'm also now wondering if we couldn't get away with all NE-dependent projects simply extending the lombiq base rules individually. What do you think?

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
Piedone: No problem, but please change the still applicable parts accordingly.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
Piedone: We need some kind of common ESLint config since we don't want to have to maintain config overrides in N places. Due to the reusable and shared nature of all OSOCE projects it's not necessarily exactly the same here, but generally in all solutions we need a central, single config for ESLint. This might be achievable in different ways though, not just what we've done before.

Piedone: Why does an upfront build needs to happen though? Since NE is referenced by the projects it's used from, it should be at the root of the dependency tree, being built first.

0liver: I've realized in the meantime, that the problematic part of the whole configuration setup is **only** with the VS ESLint integration, not with running ESLint from NE during the build itself :grin

So, yes, no upfront build should be unnecessary.

Lombiq.NodeJs.Extensions/pnpm-lock.yaml
0liver: Looks like [`core.safecrlf`](https://git-scm.com/docs/git-config#Documentation/git-config.txt-coresafecrlf) set to `false` would achieve what we need here.

But I am currently unable to test this, because the annoying change is somehow gone 😮

Piedone: Did you also close VS, clean the repo, and build after that?

0liver: Yes, all of it.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: I've got the VS ESLint support working now for JS files in the Samples project, with both locations for the _.eslintrc_ file, in a project, and inside the solution root folder.

For some reason it doesn't work yet in the NuGet samples, yet. I'm working on it.

Lombiq.NodeJs.Extensions/package.json
0liver: I'm refraining from changing this right now because the babel config file also contains settings for the sitemap generation and I wouldn't like that to be exposed to the consumer. So I'm leaving this part as is.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
```diff
@@ -40,42 +40,64 @@ To use the `npm` scripts defined in this project, add any or all of the followin
...
+## ESLint rules
+
+The rules are found in 2 files:
+- *.eslintrc.lombiq-base.js*: This file contains Lombiq overrides for the [airbnb-base](https://www.npmjs.com/package/eslint-config-airbnb-base) rules. It is located in *node_modules/nodejs-extensions/config*.
+- *.eslintrc.json*: In this file you can override the above Lombiq rules, or define your own [ESLint configuration](https://eslint.org/docs/user-guide/configuring/configuration-files) altogether.
+
+The *.eslintrc.json* file initially extends *.eslintrc.lombiq-base.js* from the Node.js Extensions `npm` package. It will automatically be created in your project during the first build. Should you prefer to use a global *.eslintrc.json* file for your whole solution, you can instruct Node.js Extensions to create that file in the location specified by the MSBuild property `<NodeJsExtensionsGlobalESLintConfigurationDirectory>`. This property is easiest added in a *Directory.Build.props* file in your solution's root directory as follows:
+
+```xml
+<NodeJsExtensionsCreateESLintConfigurationFile>$(MSBuildThisFileDirectory)</NodeJsExtensionsCreateESLintConfigurationFile>
```
Piedone: This and `NodeJsExtensionsCreateStylelintConfigurationFile` take a different approach to the same problem. Let's settle on one approach and use this in both cases (i.e. looking at Directory.Build.props, one should see something like `SomeStylelintConfigForNodJsExtensions` and `SomeESLintConfigForNodJsExtensions` working exactly in the same fashion).

Lombiq.NodeJs.Extensions/config/consumer/package.project.json
```diff
@@ -0,0 +1,16 @@
+{
+  "scripts": {
+    "build": "npm explore nodejs-extensions -- pnpm build",
+    "clean": "npm explore nodejs-extensions -- pnpm clean",
+    "watch": "npm explore nodejs-extensions -- pnpm watch"
```
Piedone, line 5: I get these errors when starting `watch` and it indeed doesn't rebuild on change (but repeats this crash):

image

Manual build works.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -54,14 +53,64 @@
...
+    <Warning
+      Code="NE01"
+      Text="The path '$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' in 'NodeJsExtensionsGlobalESLintConfigurationDirectory' is not a directory."
+      Condition="'$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != '' AND '$(IsGlobalESLintConfigurationDirectoryValid)' != 'true'" />
+    <!-- We need pnpm, so we enable it here explicitly using `corepack enable` from Node.js >=16.9. -->
+    <Exec Command="corepack enable" IgnoreExitCode="true"
+          Condition="'$(IsGlobalESLintConfigurationDirectoryValid)' == 'true'" />
+    <!-- We use the latest pnpm@6 version, which sadly cannot be expressed more succinctly.-->
+    <Exec Command="corepack prepare pnpm@6.32.24 --activate" IgnoreExitCode="true"
+          Condition="'$(IsGlobalESLintConfigurationDirectoryValid)' == 'true'" />
```
Piedone, line 97: It seems to me that these take care of the PNPM dependency and thus manual installation as per the Readme is not necessary anymore.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.json
```diff
@@ -0,0 +1,11 @@
+{
```
Piedone, line 1: Shouldn't this file and .eslintrc.project.json kept in sync? If yes, add a comment about it in both. Same for package.json files.

Lombiq.NodeJs.Extensions/scripts/minify.js
```diff
@@ -1,43 +1,60 @@
...
+            exec(command, (err, stdout, stderr) => {
+                if (err) {
+                    console.error(`${filePath}: ${err}`);
+                }
+                else {
+                    // Print the *entire* stdout and stderr (buffered).
+                    stdout && console.log(`${filePath}: ${stdout}`);
+                    stderr && console.log(`${filePath}: ${stderr}`);
+                }
+            });
```
Piedone, line 61: There are some warnings here: 
image

Piedone (changes requested):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -30,36 +34,90 @@
...
+    <!-- Remove previous version of the Node.js Extensions npm package; keeping node_modules in place, though, for
+         performance reasons. -->
+    <ItemGroup>
+      <NodeJsExtensionsNpmPackageFoldersToDelete Condition="!Exists($(NodeJsExtensionsNpmPackageSourcePath))" />
+      <NodeJsExtensionsNpmPackageFoldersToDelete
+        Include="$([System.IO.Directory]::GetDirectories($(NodeJsExtensionsNpmPackageSourcePath)))"
+        Exclude="$([System.IO.Directory]::GetDirectories($(NodeJsExtensionsNpmPackageSourcePath), 'node_modules'))"
+        Condition="Exists($(NodeJsExtensionsNpmPackageSourcePath))" />
+    </ItemGroup>
+    <RemoveDir Directories="@(NodeJsExtensionsNpmPackageFoldersToDelete)" ContinueOnError="true" />
```
Piedone: Re `ContinueOnError` see this convo, might be applicable: https://github.com/Lombiq/NPM-Targets/pull/20#discussion_r915075950

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: I agree, and this will happen as part of finishing OSOE-84. I've introduced these inconsistencies as improvements to the previous solution that I build for the SCSS pipeline. On the other hand, I felt that refactoring the latter was out of scope for the issue at hand.

Lombiq.NodeJs.Extensions/config/consumer/package.project.json
0liver: This is the breaking change I mentioned [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/pull/124#discussion_r918845962). It should work fine with **pnpm@v6**.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
0liver: Good point!

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: The consumer of NE could still be using Node.js with a version < 16.9, where `corepack` does not exist yet. In that case, these calls will fail, which will be ignored, but as long as pnpm is installed, our script should still work.

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
Piedone: Does https://lombiq.atlassian.net/browse/OSOE-84 include a note about this?

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: We can require that Node version, if it's available on GitHub Actions, there's no issue installing it in TC, and team members can do that too (https://lombiq.atlassian.net/browse/ADHOC-158).

0liver: Just because we can, doesn't mean we should.

As it currently stands, NE will be usable even for lower Node.js versions as long as `pnpm` will be available. Why would we want to break that?

Lombiq.NodeJs.Extensions/Docs/JavaScript.md
0liver: Now it does.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.json
0liver: I don't feel there's much to keep in sync for the `.eslintr.*.json` files.

I've added a comment to the `package.*.json` files, though.

Lombiq.NodeJs.Extensions/scripts/minify.js
0liver: I've ignored the `no-console` rule for the whole file, but I'm not getting the first warning, so I'm not fixing that.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: Fair enough. The Readme asks you to enable corepack though, which is not necessary.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.json
Piedone: There's not yet. But if we happen to change any of them, we'll need to remember to consider the other one too.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: I've removed that part and added the version restriction for `pnpm`.

Lombiq.NodeJs.Extensions/scripts/minify.js
Piedone: I still get "(ESLint) import/no-unresolved	Unable to resolve path to module 'klaw'." here.

Piedone (changes requested):
See https://github.com/Lombiq/NodeJs-Extensions/pull/3#discussion_r920308501

Lombiq.NodeJs.Extensions/scripts/minify.js
0liver: Well, for the life of me, I don't get the error. I've deep-cleaned the repo, and neither before nor after building NE or the solution do I see the error message you're mentioning.

Can you successfully run the scripts, though?

0liver: If yes, then let's call it a day, shall we? We didn't lint our Gulp etc. files in GE at all, so this is still an improvement.

0liver: OK, I've added a rule suppression for that line now.

Piedone (approved):

PR #7, opened 2022-07-26
OSOE-84: Replace Gulp Extensions with Node.js Extensions in OSOCE

0liver:
OSOE-84


Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,104 @@
...
+            .then(() => { log('Finished executing copy-assets.js. (1)'); })
+            .catch((inner) => process.stderr.write(JSON.stringify(inner)));
+
+        log('Finished executing copy-assets.js. (2)');
+    }
+    catch (error) {
+        process.stderr.write(JSON.stringify(error));
+    }
+
+    log('Finished executing copy-assets.js. (3)');
```
0liver: @DAud-IcI Can you spot why neither of the three log statements are displayed while those from line 88 and those from inside the `copyFilesFromConfig()` call in line 89 are? It just doesn't make any sense to me.

Also, if I inline the `main()` method, and after removal of the `await` in line 85, the log statements `(2)` and `(3)` are displayed, but still not `(1)`. What gives?

sarahelsaig: I've just checked this out and they get displayed for me:
image

> Also, if I inline the `main()` method, and after removal of the `await` in line 85, the log statements `(2)` and `(3)` are displayed, but still not `(1)`. What gives?

If you remove the `await` on line 85, then line 97 and onwards are not sequentially after the Promise any more. Since Javascript is strictly single-threaded, the way this works is that the Promise-returning method (`readConfig`) schedules an event (put it into an event queue) and then the scheduler will look at it once the current function scope is over. So at best you get (2), (3), (1). Also I'm not certain Node would wait for the Promise in this scenario since it wasn't returned.

0liver: Thanks a lot for trying it out, Dávid! I'm perplex to see the output appeared for you.

I've noticed, though, that the line that's displayed just before (1) is "No configuration found" which might just be the explanation. Looks like `return copyFilesFromConfig()` in line 89 returns something that breaks that 😕 I'll have another look.

0liver: After some more digging I now see the messages, too. 🎉

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -30,15 +30,15 @@
     <None Include="config/**" Pack="true" PackagePath="config/" />
     <None Include="scripts/**" Pack="true" PackagePath="scripts/" />
     <None Include="Stylelint/**" Pack="true" PackagePath="Stylelint/" />
-    <None Remove="node_modules/**" />
+    <None Remove="node_modules\**" />
```
sarahelsaig: Why change the slash for this but not the others?

Lombiq.NodeJs.Extensions/scripts/get-config.js
```diff
@@ -0,0 +1,58 @@
...
+    logLine(`Loaded configuration: ${JSON.stringify(nodejsExtensionsConfig)}`);
+
+    const interpolatedConfig = { ...defaults, ...nodejsExtensionsConfig };
+
+    logLine(`Interpolated configuration: ${JSON.stringify(interpolatedConfig)}`);
+
+    if (Array.isArray(interpolatedConfig.assetsToCopy)) {
+        interpolatedConfig.assetsToCopy.forEach((group) => {
+            // '**' is problematic because it also matches the given directory itself, which breaks the copyfiles tool.
+            if (!group.pattern || group.pattern === '**') group.pattern = defaultAssetsFilePattern;
```
sarahelsaig: Shouldn't we emit a log/warning in case of this problematic value?

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
...
+ *              "source": "CustomJsFolder",
+ *              "target": "wwwCustomRoot/jay-es"
+ *            },
+ *            "styles": {
+ *              "source": "NonDefaultScssFolder",
+ *              "target": "wwwwNonDefaultCssFolder"
+ *            }
+ *          }
+ */
+function logError(message) { process.stderr.write(`ERROR: ${message}\n`); }
```
sarahelsaig, line 30: Once OSOE-92 is merged, please replace this with `const { handleErrorObject } = require('./handle-error');`. Same with the other `process.stderr.write` in get-path.js and get-config.js.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
...
+        if (assetGroup.pattern && typeof assetGroup.pattern !== 'string') {
+            errors.push('pattern must be a glob string');
+        }
+        if (typeof assetGroup.target !== 'string') {
+            errors.push('target must be a string');
+        }
+        if (errors.length > 0) {
+            logError(`Invalid asset group: ${JSON.stringify(assetGroup)}: ${errors.join(', ')}.`);
+        }
+    });
```
sarahelsaig: I think the `const errors = [];` should be inside the `.forEach()`. Right now errors accumulate so if two `assetGroup` has errors, the first one's errors will appear for the the second group too.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
...
+function validateAssetGroupsAndLogErrors(assetConfig) {
+    if (!Array.isArray(assetConfig)) {
+        logError('The configuration must be an array of asset groups of the form: ' +
+            '{ sources: Array<string>, pattern: string [optional], target: string }.');
+        return false;
+    }
+
+    const errors = [];
+
+    assetConfig.forEach((assetGroup) => {
```
sarahelsaig: Instead of `assetConfig.forEach` you can do `return assetConfig.reduce((success, assetGroup) => {  }, true)`. In the callback return `success` if there were no errors or `false` if there were. This way you don't need to maintain an external variable and you can return immediately.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
...
+}
+
+function validateSimpleGroupAndLogErrors(group) {
+    const errors = [];
+    if (typeof group.source !== 'string') {
+        errors.push('source must be a string');
+    }
+    if (typeof group.target !== 'string') {
+        errors.push('target must be a string');
+    }
```
sarahelsaig: These can be one-liners.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
...
+    if (errors.length > 0) {
+        logError(`Invalid group: ${JSON.stringify(group)}: ${errors.join(', ')}.`);
+    }
+    return errors.length === 0;
+}
+
+function validateAndLogErrors(nodejsExtensionsConfig) {
+    return Object.entries(nodejsExtensionsConfig).every(([key, group]) => {
```
sarahelsaig, line 68: I didn't realize you can destructure an array in the function parameter. This is amazing!

Lombiq.NodeJs.Extensions/scripts/validate-config.js
```diff
@@ -0,0 +1,89 @@
+/**
+ * @summary Validates the Node.js Extensions configuration object.
```
sarahelsaig, line 2: Not important for this PR, but isn't this basically schema validation? We should consider replacing the custom logic with existing tools. This looks suitable: https://www.npmjs.com/package/jsonschema If you agree I'll create a followup issue.

sarahelsaig (changes requested):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
0liver: Yeah, why?

Lombiq.NodeJs.Extensions/scripts/get-config.js
0liver: We could, that's true.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
0liver: Good catch!

Lombiq.NodeJs.Extensions/scripts/validate-config.js
0liver: I completely agree, but I wanted to avoid scope explosion. Go right ahead!

Lombiq.NodeJs.Extensions/scripts/validate-config.js
0liver: Beautiful!

Lombiq.NodeJs.Extensions/scripts/validate-config.js
0liver: I had them like this in conformation with the method above this one, but removing 4 unneeded lines of code actually felt great.

Lombiq.NodeJs.Extensions/scripts/validate-config.js
sarahelsaig: I've created the issue: https://github.com/Lombiq/NodeJs-Extensions/issues/15

I'm all for standing up to scope explosion. We should be more willing to create followup issues instead of bloating existing ones.

PR #8, opened 2022-07-27
WALMA-4: Fixing package.json errors

Luko6:
WALMA-4


Lombiq.NodeJs.Extensions/config/consumer/package.global.json
```diff
@@ -1,5 +1,4 @@
 {
-  // Keep the eslint* entries in 'devDependencies' in sync with those in package.json.
```
0liver, line 2: Please add this line to a _NOTES.md_ file in the same folder.

dministro (approved):

PR #10, opened 2022-07-29
OSOE-161: Add Copy Assets functionality

0liver:
OSOE-161


Lombiq.NodeJs.Extensions/Docs/AssetCopying.md
```diff
@@ -0,0 +1,42 @@
...
+
+Should you prefer to use a different target path, please adjust it in your project file by adding the following property:
+
+```xml
+<NodeJsExtensionsAssetsTargetFolder>wwwroot/other</NodeJsExtensionsAssetsTargetFolder>
+```
+
+## Configuration
+
+You need to provide a map of source and target paths either in a separate file called _assets-to-copy.json_ or as a property in your _package.json_ named `assetsToCopy`. The configuration will be loaded in that order. Its format is as follows:
```
sarahelsaig: Seems superfluous to me.
```suggestion
You need to provide a map of source and target paths either in a separate file called _assets-to-copy.json_ or as a property in your _package.json_ named `assetsToCopy`. The configuration will be loaded in that order. Its format is:
```

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -31,11 +34,24 @@
     <UpToDateCheckInput Include="@(NpmDotnetPrebuildWatchedFiles)" />
   </ItemGroup>
 
+  <!-- We need to install the dependencies of the currently built project for asset copying to work (at least). -->
```
sarahelsaig: What does "(at least)" mean here?

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -15,6 +15,8 @@
       SCSS_SOURCE=$(NodeJsExtensionsStylesSourceFolder.Replace('\', '/'));
       SCSS_TARGET=$(NodeJsExtensionsStylesTargetFolder.Replace('\', '/'));
     </NpmCommandsEnvironmentVariables>
+    <!-- The importance with which to print informational messages. -->
+    <Importance>High</Importance>
```
sarahelsaig: Shouldn't this be
```suggestion
    <Importance Condition="'$(Importance)' = ''">High</Importance>
```
so it's only set to "High" if the user didn't already provide a value in the csproj?

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
```diff
@@ -0,0 +1,43 @@
...
+const util = require('util');
+const rimraf = util.promisify(require('rimraf'));
+const getAssetsConfig = require('./get-assets-config');
+
+const verbose = true;
+const filePattern = '**/*';
+
+// Change to consuming project's directory.
+process.chdir('..');
+process.chdir('..');
```
sarahelsaig: Is this meant to be twice? If so, use `process.chdir('../..')` instead. This works on Windows to if you were worried about that. You can also use `path.resolve('..', '..')` instead of `'../..'`.
Same [here](https://github.com/Lombiq/NodeJs-Extensions/blob/44782f9ed55eecba85faa109ea88b75f68d0bda7/Lombiq.NodeJs.Extensions/scripts/copy-assets.js#L17-L18).

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
```diff
@@ -0,0 +1,43 @@
...
+
+function log(message) {
+    if (verbose) process.stdout.write(message + '\n');
+}
+
+async function deleteFiles(assetsConfig) {
+    log('Executing deleteFiles ...');
```
sarahelsaig:
```suggestion
    log('Executing deleteFiles...');
```

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
```diff
@@ -0,0 +1,43 @@
...
+
+    return Promise.all(assetsConfig
+        .map((assetsGroup) => {
+            log(`Cleaning ${assetsGroup.target}`);
+
+            return rimraf(path.join(assetsGroup.target, filePattern));
```
sarahelsaig, line 27: Whoever came up with this name will be punished... severely. 😒

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
```diff
@@ -0,0 +1,43 @@
...
+        }));
+}
+
+(async function main() {
+    log('Executing clean-assets.js ...');
```
sarahelsaig: Why do you add spaces in front of the punctuation?
If you want to delineate the file name, rather wrap it in quotes.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,60 @@
+/**
+ * @summary A script to copy files around; expects configuration in assets-to-copy.json or package.json (assetsToCopy).
+ * @description This script is a wrapper around the npm package "copyfiles". It reads its configuration from a file in
+ *              the consuming project named "assets-to-copy.json" or the "assetsToCopy" node in package.json.
+ */
+const { access } = require('node:fs/promises');
```
sarahelsaig: I got complaints from Zoltán when I used `node:fs/promises`, because it's apparently not in v14. You can use `require('fs').promises` instead (like I did [in this commit](https://github.com/Lombiq/NodeJs-Extensions/commit/82a020ab056b96afb383cda4470ef9a27d72f7c3)) so let's use that for now. (same for all other `node:fs/promises` uses)

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,60 @@
...
+const path = require('path');
+const util = require('util');
+/* eslint-disable-next-line import/no-unresolved -- ESLint does not know where to find external modules; ignore. */
```
sarahelsaig: Is the `; ignore` useful? `eslint-disable` already communicates that, I think.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,60 @@
...
+const copyfiles = util.promisify(require('copyfiles'));
+const getAssetsConfig = require('./get-assets-config');
+
+const verbose = true;
+const filePattern = '**/*';
```
sarahelsaig: Isn't that only the default value? Name should indicate that.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,60 @@
...
+
+async function copyFilesFromConfig(assetsConfig) {
+    return Promise.all(assetsConfig
+        .map((assetsGroup) => assetsGroup.sources.map((assetSource) => {
+            // Normalize the relative path to the directory to remove trailing slashes and straighten out any anomalies.
+            const directoryToCopy = path.relative('.', assetSource);
+            log(`directoryToCopy = ${directoryToCopy}`);
+
+            return access(directoryToCopy)
+                .catch(() => process.stderr.write(`The directory "${directoryToCopy}" cannot be accessed to copy files from.\n`))
```
sarahelsaig: Shouldn't this use MSBuild error format?

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,60 @@
...
+                .then(() => {
```
sarahelsaig: Merge these two together (instead of `.catch(f1).then(f2)` it should be `.then(f2, f1)`). Why? Because a resolved `.catch` is still a successful promise. So if you can't access the directory it will write out the message and then try to access the directory with the `.then` anyway.

You can test this by typing
```javascript
Promise.reject('expected')
    .catch(() => 'unexpected')
    .then((result) => console.log(result)); 
```
It will print "unexpected".

Using `.then(resolveCallback, rejectCallback)` format avoids this. Alternatively, you can move the `.catch` underneath the `.then`. Then any failure from either the `access()` or the `.then` will be handled by it.

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
```diff
@@ -0,0 +1,90 @@
...
+        })
+        .then((config) => {
+            logn(config ? `Loaded configuration: ${JSON.stringify(config)}` : 'No configuration found.');
+            if (config) {
+                if (!Array.isArray(config)) {
+                    logError('The configuration must be an array of asset groups of the form: ' +
+                        '{ sources: Array<string>, pattern: string [optional], target: string }.');
+                    return null;
+                }
+                // Return at least all valid assets groups for some user satisfaction.
```
sarahelsaig: Is "some" satisfaction enough? Would it be more reliable to `process.exit(1)` out if any is invalid.

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
```diff
@@ -0,0 +1,90 @@
...
+    }
+    if (errors.length > 0) {
+        logError(`Invalid asset group: ${JSON.stringify(assetsGroup)}: ${errors.join(', ')}.`);
+    }
+    return errors.length === 0;
+}
+
+async function getAssetsConfig({ directory, verbose } = { directory: process.cwd(), verbose: false }) {
+    const log = (message) => { if (verbose) process.stdout.write(message); };
+    const logn = (message) => { log(message + '\n'); };
```
sarahelsaig: `logn` sounds like something mathematicsy. Also you don't need the curly braces.

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
```diff
@@ -0,0 +1,90 @@
...
+
+    const assetsJsonPath = path.resolve(directory, assetsFileName);
+    const packageJsonPath = path.resolve(directory, 'package.json');
+
+    return Promise.resolve()
+        .then(() => log(`Reading configuration from ${assetsJsonPath} ... `))
```
sarahelsaig: why log and not logn?

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
```diff
@@ -0,0 +1,90 @@
...
+        .then(() => readFile(assetsJsonPath, 'utf-8'))
```
sarahelsaig: Why? You can have the first log outside of the promise, and start with `readFile` so you don't need the `Promise.resolve()`.

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
```diff
@@ -0,0 +1,19 @@
...
+(async function main() {
+    try {
+        const stylesSourceDirectory = path.resolve('..', '..', process.env.SCSS_SOURCE);
+        await access(stylesSourceDirectory);
+        process.exit(1);
+    }
+    catch (_) {
+        process.stdout.write('Skipping styles');
+        process.exit(0);
+    }
```
sarahelsaig: Why are you returning 1 on success and 0 on failure? This is very abnormal.

Readme.md
```diff
@@ -60,7 +60,8 @@ Here's an overview of all of the scripts this project makes available, categoriz
 
 1. [Styles](Lombiq.NodeJs.Extensions/Docs/Styles.md)
 2. [JavaScript](Lombiq.NodeJs.Extensions/Docs/JavaScript.md)
-3. [Markdown](Lombiq.NodeJs.Extensions/Docs/Markdown.md)
+3. [Asset Copying](Lombiq.NodeJs.Extensions/Docs/AssetCopying.md)
+4. [Markdown](Lombiq.NodeJs.Extensions/Docs/Markdown.md)
```
sarahelsaig: Why did you insert Asset Copying in the middle of the list? Does this order represent something?
I think this list doesn't need to be numbered, and it could be just sorted alphabetically. What is your opinion?

sarahelsaig (changes requested):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
0liver: image

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
0liver: I'll add an underscore to convey that this is an internal flag. And I'll set it to `Normal`, too. Useful for debugging, but not much else.

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
0liver: I felt like avoiding forward slashes in paths, if we're already using Node. And I didn't want to include `path` just for that. So I thought that the solution I came up with was both clear, platform-independent, and without needing another dependency.

Is it ugly to you?

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
0liver: I didn't like how that looked, but I can save that space here and there.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
0liver: Thanks, that made it crystal clear. 😎

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
sarahelsaig: Not just ugly, but more importantly if I see two identical lines one after the other I get suspicious that it's maybe an accidental duplication. Worse yet, someone might act on that suspicion.

Also why are you against using forward slashes in Node paths?  Like I said, works perfectly fine in Windows.
image

Also also, there is no cost associated with including `path` because as a system dependency. It's loaded in by default:
image
See? It was already there without needing a `require`. The same as `process`. (I mean if you want you are technically allowed to do `const process = require('process');` but you didn't.)

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
0liver: You're probably right.

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
0liver: I loved how fluent this is!

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
0liver: Because I'm checking whether I should **skip** processing **styles**.

Readme.md
0liver: I was thinking that the first three items refer to processing assets in a project for publishing, while 4. does static analysis. So I wanted to group them according to that.

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
0liver: I've chosen to name the NL terminated versions `logLine`.

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
0liver: Convinced 😅 Thanks again for the insights.

Readme.md
sarahelsaig: That seems arbitrary. You could make it explicit with a sublist or something.

Also why are publishing items in the Styles, JavaScript, Asset Copying order?

0liver: I've come to like your suggestion of using an unordered, alphabetically sorted list.

0liver: What a beautiful staircase:

```markdown
- [Asset Copying](Lombiq.NodeJs.Extensions/Docs/AssetCopying.md)
- [JavaScript](Lombiq.NodeJs.Extensions/Docs/JavaScript.md)
- [Markdown](Lombiq.NodeJs.Extensions/Docs/Markdown.md)
- [Styles](Lombiq.NodeJs.Extensions/Docs/Styles.md)
```

😁

sarahelsaig: Ah, so that's what they meant by "waterfall model"! 🙃

Lombiq.NodeJs.Extensions/scripts/clean-assets.js
0liver: Fixed.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -0,0 +1,63 @@
...
+        .reduce((previousArray, currentArray) => [...previousArray, ...currentArray], []));
+}
+
+(async function main() {
+    logLine('Executing copy-assets.js...');
+
+    try {
+        await getAssetsConfig({ directory: process.cwd(), verbose: verbose })
+            .then((config) => (config ? copyFilesFromConfig(config) : Promise.resolve()))
+            .then(() => { logLine('Finished executing copy-assets.js.'); });
```
sarahelsaig: It just occurred to me, but why are you mixing async and then chains?

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
sarahelsaig: Using a bunch of `.then()`s doesn't seem to respect the spirit of a fluent interface.
This is just method chaining for aesthetic reasons.

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
sarahelsaig: It's still incredibly alien and unnatural to return failure code in the try and success code in the catch.

What's wrong with `const styleDirectoryExists = fs.existsSync(path.resolve('..', '..', process.env.SCSS_SOURCE))` that returns Boolean instead of throws?

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
sarahelsaig: Or how about making the function `async` and using `await`s instead?

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
0liver: Checking the existence of a path and verifying access to it is not the same. Maybe this is me being overly cautious but when I read about `fs.access` it actually made a lot of sense to use that instead of `fs.exists`, because we'll actually want to write to the path in question.

The other decision I made was to use `access` over `accessSync` because I thought that's preferable because the executing thread doesn't block on the I/O operation. And that's how I ended up with this "incredibly alien and unnatural" piece of code.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
0liver: Probably because I'm still learning these concepts and don't have any experience in writing `async` code. But I've also found code samples that use both together, so it didn't feel that bad.

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
sarahelsaig: Currently if the path is inaccessible it will skip the step, failing silently. I don't think this is a desirable behavior.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
sarahelsaig: Maybe they had some good reason, but here it's unnecessary complication. You can replace the wrapping try-catch with a final `.catch()` at the end of the chain and replace the `await` with a `return`. Or you can await each of these 3 lines individually. The end result is the same.

Lombiq.NodeJs.Extensions/scripts/skip-styles.js
0liver: That's a very good point!

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
0liver: Much nicer!

Lombiq.NodeJs.Extensions/scripts/get-assets-config.js
0liver: As to why... so I can add the status at the end of the same line:

> 5>[copy:assets  ] Executing copy-assets.js...
5>[copy:assets  ] Reading configuration from C:\Projects\Lombiq\OSOCE\src\Modules\Lombiq.ChartJs\Lombiq.ChartJs\assets-to-copy.json... **failed**.
5>[copy:assets  ] Reading configuration from C:\Projects\Lombiq\OSOCE\src\Modules\Lombiq.ChartJs\Lombiq.ChartJs\package.json... **succeeded**.

Readme.md
sarahelsaig: You said you liked my suggestion, but then didn't commit it?

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
sarahelsaig: You have the same in _clean-assets.js_ too.

https://github.com/Lombiq/NodeJs-Extensions/blob/52549353b70d4eee3daa87cd890228ac37374883/Lombiq.NodeJs.Extensions/scripts/clean-assets.js#L35-L37

Readme.md
0liver: Indeed, for some reason it ended up in a stash 😏

PR #12, opened 2022-08-08
OCC-15: Add pre-requisites section.

sarahelsaig:
[OCC-15](https://lombiq.atlassian.net/browse/OCC-15)


Piedone (approved):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -106,7 +106,8 @@
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/.eslintrc.project.json"
           DestinationFiles="$(MSBuildProjectDirectory)/.eslintrc.json"
           Condition="
-            !Exists($(NodeJsExtensionsGlobalESLintConfigurationDirectory))
+            '$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != ''
```
0liver: Unfortunately, this was too quick of a shot - because with that change the default behavior, where you don't overwrite anything, will not work anymore.

I think it would be good for NE opt out of creating the file in case where the Scripts pipeline won't be executed due to missing input, i.e. non-existent folder or files in them.

But for the scope of this issue, please use `'false'` as the discriminating value instead of `''` because that at least won't break the default, project based scenario. Update the value in the OCC project file accordingly.

I apologize for the inconvenience.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -106,7 +106,8 @@
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/.eslintrc.project.json"
           DestinationFiles="$(MSBuildProjectDirectory)/.eslintrc.json"
           Condition="
-            !Exists($(NodeJsExtensionsGlobalESLintConfigurationDirectory))
+            '$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != 'false'
+            AND !Exists($(NodeJsExtensionsGlobalESLintConfigurationDirectory))
```
Piedone, line 110: Something went off here.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -68,13 +68,16 @@
   <Target Name="SetUpGlobalESLintConfiguration" BeforeTargets="BeforeResolveReferences">
     <PropertyGroup>
       <NodeJsExtensionsGlobalESLintConfigurationDirectory
-        Condition="'$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != ''
+        Condition="'$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' == ''
```
Piedone, line 71: OrchardCore.Commerce\src\OrchardCore.Commerce.Web\.eslintrc.json is created for me upon rebuild.

Piedone (changes requested):

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
sarahelsaig: It's not created for me. OrchardCore.Commerce\src\Modules\OrchardCore.Commerce.eslintrc.json isn't created either.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
```diff
@@ -68,13 +68,16 @@
...
+      ">$(ProjectDir)</NodeJsExtensionsGlobalESLintConfigurationDirectory>
+      <NodeJsExtensionsGlobalESLintConfigurationDirectory
+        Condition="'$(NodeJsExtensionsGlobalESLintConfigurationDirectory)' != 'false'
       ">$(NodeJsExtensionsGlobalESLintConfigurationDirectory.TrimEnd('/\'))</NodeJsExtensionsGlobalESLintConfigurationDirectory>
```
sarahelsaig: On second thought @0liver, why is TrimEnd necessary here? I think if you have an extra slash in the path that won't actually hurt. I think the TrimEnd should be removed

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.csproj
Piedone: With the latest OCC (37d8aa2798dc8741dd80ed19993f4cec218f4c86) it's good for me too.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
sarahelsaig: what do you mean?

Piedone: When I checked the diff, it was displayed as something along the lines of 

```xml
    </Target>
         AND !Exists($(NodeJsExtensionsGlobalESLintConfigurationDirectory))
```

Like a bad merge. But not it looks good.

PR #20, opened 2022-12-30
OSOE-405: Improve and refactor Markdown linting pipeline

### LINKED ISSUE #14: Improve and refactor Markdown linting pipeline (OSOE-405)
After OSOE-92 has been merged.

### Problem

Currently, _lint-markdown.js_ uses `markdownlint.sync()` to lint markdown files.

The release notes for [version 0.21.0](https://github.com/DavidAnson/markdownlint#:~:text=lint%20concurrently%20for%20better%20performance%20(async%20only)) of `markdownlint` mention:
> Lint concurrently for better performance (async only)

Let's use that.

Another reason to use the async version is that it takes a [callback](https://github.com/DavidAnson/markdownlint#callback) that receives both **Error** and **Result** objects. In the error case, we can then use `handleError` the same way that we do for `textlint`.

### Additional improvements
- Use `handleErrorMessage` from _handle-error,.js_ on `throw new Error("Couldn't find a .NET solution (.sln) file.");`.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-405)

sarahelsaig:
[OSOE-405](https://lombiq.atlassian.net/browse/OSOE-405)

Fixes #14

[OSOE-405]: https://lombiq.atlassian.net/browse/OSOE-405?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -33,6 +33,8 @@
           Inputs="$(MSBuildProjectDirectory)/package.json"
           Outputs="$(MSBuildProjectDirectory)/node_modules/.nx_pnpm_install_stamp"
           Condition="'$(NodeJsExtensionsShouldDoWork)' == 'true'">
+
+    <MakeDir Directories="$(NodeJsExtensionsDefaultTargetDirectory)" Condition="!Exists('$(NodeJsExtensionsDefaultTargetDirectory)')" />
```
0liver: Why do you need that?

sarahelsaig: I don't.

0liver (approved):

PR #24, opened 2023-02-05
OSOE-553: Automatically install the package.global.json dependencies

sarahelsaig:
[OSOE-553](https://lombiq.atlassian.net/browse/OSOE-553)

[OSOE-553]: https://lombiq.atlassian.net/browse/OSOE-553?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
+++ b/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
@@ -117,7 +117,7 @@
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/package.project.json"
           DestinationFiles="$(MSBuildProjectDirectory)/package.json"
           Condition="'$(ShouldConfigureESLintPerProject)' == 'true' AND !Exists('$(MSBuildProjectDirectory)/package.json')" />
-    <!-- Add the necessary ESLint plugin package depdendencies to the package.json file. -->
+    <!-- Add the necessary ESLint plugin package dependencies to the package.json file. -->
     <Exec Command="node &quot;$(NodeJsExtensionsNpmPackageSourcePath)/scripts/add-dev-dependencies.js&quot;"
           WorkingDirectory="$(MSBuildProjectDirectory)"
           Condition="'$(ShouldConfigureESLintPerProject)' == 'true'" />
--- a/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
+++ b/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
@@ -120,7 +120,7 @@
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/package.minimal.json"
           DestinationFiles="$(MSBuildProjectDirectory)/package.json"
           Condition="'$(ShouldConfigureESLintPerProject)' != 'true' AND !Exists('$(MSBuildProjectDirectory)/package.json')" />
-    <Exec Command="node &quot;$(NodeJsExtensionsNpmPackageSourcePath)/scripts/install-dev-dependencies.js&quot;"
+    <Exec Command="node &quot;$(NodeJsExtensionsNpmPackageSourcePath)/scripts/add-dev-dependencies.js&quot;"
           WorkingDirectory="$(MSBuildProjectDirectory)" />
   </Target>
 
--- a/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
+++ b/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
@@ -113,15 +113,14 @@
           Condition="'$(ShouldConfigureESLintPerProject)' == 'true' AND '@(ESLintrcFile)' == ''" />
     <!-- Here we link to Node.js Extensions as an npm package, which makes it available to the consumer project under
          its package name "nodejs-extensions". This needs the consumer to have a package.json file, so if there is none,
-         we create a minimal one. -->
+         we create one. -->
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/package.project.json"
           DestinationFiles="$(MSBuildProjectDirectory)/package.json"
           Condition="'$(ShouldConfigureESLintPerProject)' == 'true' AND !Exists('$(MSBuildProjectDirectory)/package.json')" />
-    <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/package.minimal.json"
-          DestinationFiles="$(MSBuildProjectDirectory)/package.json"
-          Condition="'$(ShouldConfigureESLintPerProject)' != 'true' AND !Exists('$(MSBuildProjectDirectory)/package.json')" />
+    <!-- Add the necessary ESLint plugin package depdendencies to the package.json file. -->
     <Exec Command="node &quot;$(NodeJsExtensionsNpmPackageSourcePath)/scripts/add-dev-dependencies.js&quot;"
-          WorkingDirectory="$(MSBuildProjectDirectory)" />
+          WorkingDirectory="$(MSBuildProjectDirectory)"
+          Condition="'$(ShouldConfigureESLintPerProject)' == 'true'" />
   </Target>
 
   <!--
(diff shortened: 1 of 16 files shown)


Lombiq.NodeJs.Extensions/scripts/install-dev-dependencies.js
```diff
@@ -0,0 +1,24 @@
...
+const packages = Object
+    .entries(devDependencies)
+    .map(([name, version]) => `${name}@${version}`)
+    .join(' ');
+
+// Ensure each of the default development dependencies are installed. If they already are, this finishes very quickly.
+// It is necessary, because by default NPM and PNPM don't install the dependencies of local packages, see here:
+// https://docs.npmjs.com/cli/v9/configuring-npm/package-json#local-paths
+try {
+    execSync('pnpm install --save-dev ' + packages);
```
0liver: How about we don't touch the _package.json_ file, and just install the dependencies silently, i.e. remove the `--save-dev` flag?

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -120,6 +120,9 @@
     <Copy SourceFiles="$(NodeJsExtensionsNpmPackageSourcePath)/config/consumer/package.minimal.json"
           DestinationFiles="$(MSBuildProjectDirectory)/package.json"
           Condition="'$(ShouldConfigureESLintPerProject)' != 'true' AND !Exists('$(MSBuildProjectDirectory)/package.json')" />
+    <Exec Command="node &quot;$(NodeJsExtensionsNpmPackageSourcePath)/scripts/install-dev-dependencies.js&quot;"
+          WorkingDirectory="$(MSBuildProjectDirectory)" />
+
```
0liver: Remove the empty line.

Lombiq.NodeJs.Extensions/config/consumer/package.project.json
```diff
@@ -4,13 +4,5 @@
...
     "clean": "npm explore nodejs-extensions -- pnpm clean",
     "watch": "npm explore nodejs-extensions -- pnpm watch"
-  },
-  "devDependencies": {
-    "eslint": "^8.25.0",
-    "eslint-config-airbnb-base": "^15.0.0",
-    "eslint-plugin-import": "^2.26.0",
-    "eslint-plugin-node": "^11.1.0",
-    "eslint-plugin-only-warn": "^1.0.3",
-    "eslint-plugin-promise": "^6.1.0"
```
0liver, line 14: Removing these is a good idea. I suggest removing it from _package.global.json_ as well, and reading the entries from NE's own _package.json_ file with a filter on `name.startsWith('eslint')`.

Lombiq.NodeJs.Extensions/scripts/install-dev-dependencies.js
```diff
@@ -0,0 +1,24 @@
+const fs = require('fs');
+const path = require('path');
+const { execSync } = require('child_process');
+
+const { handleErrorObject } = require('./handle-error');
+
+const configPath = path.resolve(__dirname, '..', 'config', 'consumer', 'package.global.json');
```
0liver: See my comment here: https://github.com/Lombiq/NodeJs-Extensions/pull/24/files#diff-3fb050e9d28f89a9034af1a63dfa858a6114c61db89a7726e2167b24cfea676fL8.

0liver (changes requested):

Lombiq.NodeJs.Extensions/scripts/install-dev-dependencies.js
sarahelsaig: If you remove the `--save-dev` flag it will still touch the package.json, just add them to the `dependencies` instead of the `devDependencies`. I think this is more appropriate since these are only used for linting.

0liver: Concluding from [this open PNPM issue](https://github.com/pnpm/pnpm/issues/2020), there's currently no way to avoid adding installed packages to the _package.json_ and _pnpm-lock.yaml_ 😔

0liver: I've looked into this some more, and I think we should be using `pnpm add` here instead of install. This also makes sense in the light of us calling `pnpm install` as the next step after this one. I've already successfully tried that out on my machine.

With that, I suggest to rename the script to `add-dev-dependencies.js`.

0liver (changes requested):

Lombiq.NodeJs.Extensions/scripts/install-dev-dependencies.js
0liver: I've already done that in the accompanying [suggestions PR](https://github.com/Lombiq/NodeJs-Extensions/pull/25/files).

0liver: There's another nice benefit of using `pnpm add` - it uses PNPM's default version selector, which is `^` 😁 and results in having the same package `name@version` entries that we have in our NE _package.json_.

0liver (approved):

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -3,7 +3,15 @@
  */
 
 const os = require('os');
-const chalk = require('chalk');
+
+// Treat this dependency as optional because it's not available everywhere.
```
0liver, line 7: > it's not available everywhere

What exactly does that mean?

0liver (approved):

Lombiq.NodeJs.Extensions/scripts/handle-error.js
sarahelsaig: It wasn't available when executing from the solution root via _CreateGlobalConfigFiles.targets_. I'm not sure why. This is the sole external dependency of the script and it's only cosmetic, so in principle it's nicer to make it optional.

But I don't understand why we need this package at all. MSBuild should color and format the output by itself when the terminal supports it. Also on GHA the MSBuild error/warning is captured and transformed into an annotation.

0liver: This is for coloring output in the VS Task Explorer Runner, which didn't show any colors otherwise.

Can you point me to the erroneous run? Windows or Linux? I'm asking because I haven't had any problems with this in the past. (Maybe, I didn't have any errors to report, either, though.)

sarahelsaig: I had it on Linux, but that's probably because [the error](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/4143907294/jobs/7166322503) only appeared on Linux as it was related to the path used ([see commit](https://github.com/Lombiq/NodeJs-Extensions/pull/24/commits/f25c11cab070b60d804f081804e05d75cfce91c2)).

0liver: I'm hesitant to keeping the change, because I don't see any trace of `chalk` causing any problems looking at the very detailed binary log:
image

There's not a single mention of `chalk` in the whole log.

sarahelsaig: Dunno, it was days ago. But it's pretty easy to reproduce it even on Windows:
image
If you don't know a quick fix for this, then it would be better to open a new issue as it shouldn't be a blocker.

Also note that I will absolutely use _handle-error.js_ wherever it makes sense, not just inside NE scripts (e.g. it's in LV too). I'd rather it not have required dependencies in error handling because that could break things silently.

0liver: I've run `dotnet build` after replacing `pnpm add` with `pnpm addy`, and did not see any problem with chalk missing:

image

So I feel, that the change in code here is unnecessary.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -3,7 +3,15 @@
...
+let chalk;
+try {
+    chalk = require('chalk');
+}
+catch {
+    chalk = false;
+}
```
0liver, line 14: Looks to me like we don't need to carry this around.

0liver (changes requested):

Lombiq.NodeJs.Extensions/scripts/handle-error.js
0liver: I can agree that the dependency might not be necessary for the core _handle-error_ script, so I suggest to move the coloring of the error into an optional callback parameter.

But in a new issue.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
0liver: https://github.com/Lombiq/NodeJs-Extensions/pull/24#discussion_r1104545385

Lombiq.NodeJs.Extensions/scripts/handle-error.js
0liver: Here it is: https://github.com/Lombiq/NodeJs-Extensions/issues/30.

PR #32, opened 2023-02-17
OSOE-476: Extract inline javascript block into file, and register for usage in Lombiq.HelpfulExtensions

TheHydes:
[OSOE-476](https://lombiq.atlassian.net/browse/OSOE-476)

[OSOE-476]: https://lombiq.atlassian.net/browse/OSOE-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions/scripts/get-path.js
+++ b/Lombiq.NodeJs.Extensions/scripts/get-path.js
@@ -9,7 +9,7 @@ const process = require('process');
 const getCwd = require('./get-cwd');
 const getConfig = require('./get-config');
 
-const verbose = true;
+const verbose = false;
 const solutionFolderMarker = '_solution_';
 const extensionToTypeMap = {
     js: 'scripts',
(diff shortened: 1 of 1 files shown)


Lombiq.NodeJs.Extensions/scripts/get-path.js
```diff
@@ -9,7 +9,7 @@ const process = require('process');
 const getCwd = require('./get-cwd');
 const getConfig = require('./get-config');
 
-const verbose = false;
+const verbose = true;
```
DemeSzabolcs: This was just for debugging if I know it right. So you shouldn't need this (including this PR).

DemeSzabolcs (changes requested):

DemeSzabolcs:
We don't need this PR, the code change was just for testing and it's already reverted.

PR #33, opened 2023-02-17
OSOE-559: Make pnpm retries more robust to prevent errors

0liver:

### DIFF
--- a/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
+++ b/Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
@@ -27,7 +27,6 @@
     <!-- Stamp files to enable incremental builds. -->
     <PnpmInstallStamp>$(MSBuildProjectDirectory)/node_modules/.nx_pnpm_install_stamp</PnpmInstallStamp>
     <PnpmLinkStamp>$(MSBuildProjectDirectory)/node_modules/.nx_pnpm_link_stamp</PnpmLinkStamp>
-    <PnpmLinkCommand>pnpm link "$(NodeJsExtensionsNpmPackageSourcePath)"</PnpmLinkCommand>
   </PropertyGroup>
 
   <!-- We need to install the project's npm packages first, so that the asset copy pipeline can actually copy them. -->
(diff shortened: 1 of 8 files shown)


Lombiq.NodeJs.Extensions/build/Invoke-Sequentially.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+Write-Output "$MessagePrefix Acquiring lock..."
+
+$timeoutMs = 20000
+$sleepMs = 100
+$maxTries = $timeoutMs / $sleepMs
+$stream = $null
+
+# Try to get an exclusive lock on a lock file to guarantee only one thread will execute the given $Command at any given time.
+for ($i = 0; $i -lt $maxTries -and $null -eq $stream; $i++)
```
dministro: Use meaningful names.
```suggestion
for ($currentTry = 0; $currentTry -lt $maxTries -and $null -eq $stream; $i++)
```

Lombiq.NodeJs.Extensions/build/Invoke-Sequentially.ps1
```diff
@@ -0,0 +1,56 @@
...
+    catch [System.IO.IOException]
+    {
+        # These are expected to happen, keep going.
+        Start-Sleep -Milliseconds $sleepMs
+    }
+    catch
+    {
+        # When an unexpected exception is thrown, fail the script.
+        Write-Output $PSItem.Exception
+        exit 1
```
dministro, line 40:
```suggestion
        # The finally block runs if an exit keyword stops the script. See: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_try_catch_finally?view=powershell-7.3#freeing-resources-using-finally
        exit 1
```

dministro (changes requested):
Please mention powershell in the pre-requirements in the docs (Windows, Linux).

Lombiq.NodeJs.Extensions/build/Invoke-Sequentially.ps1
0liver: I don't think we need to add this comment here, as this is the normal behavior of a finally block and should be expected. Or do you disagree?

dministro: This is just a suggestion but was not clear to me, so this is why I suggested this comment.

0liver: I see. I would still argue that we shouldn't comment on built-in language features.

In this concrete case, the `finally` block exist for exactly the purpose of **always** running, whatever happens inside the the `try` and `catch` blocks.

Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -156,6 +169,7 @@
     <ItemGroup>
       <NodeJsExtensionsTargetFiles
         Include="%(NodeJsExtensionsTargetDirectories.Identity)/**/*"
+        Exclude="%(NodeJsExtensionsTargetDirectories.Identity)/**/_temp_/**/*"
```
0liver, line 172: I've added this to fix an intermittent error that sometimes showed up, and it happened to me again, today:

image

Lombiq.NodeJs.Extensions/build/Invoke-Sequentially.ps1
0liver: You can resolve this conversation then.

Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets
```diff
@@ -29,39 +29,16 @@
...
-      <Output TaskParameter="ExitCode" PropertyName="PnpmInstallExitCode" />
-    </Exec>
-    <!-- Use a random back-off <=10s before retrying a failed call to prevent further concurrent calls. -->
-    <PropertyGroup>
-      <TimeoutSeconds>$([MSBuild]::Modulo($([System.DateTime]::Now.Millisecond), 10))</TimeoutSeconds>
-      <Sleep>ping -n $(TimeoutSeconds) 127.0.0.1 >NUL</Sleep>
-    </PropertyGroup>
-    <Exec Command="$(Sleep)" Condition="'$(PnpmInstallExitCode)' != '0'" IgnoreExitCode="true" />
-    <!-- The last try. -->
-    <Exec Command="$(PnpmInstallCommand)" WorkingDirectory="$(NodeJsExtensionsNpmPackageSourcePath)" Condition="'$(PnpmInstallExitCode)' != '0'"/>
```
0liver, line 64: All of this has been replaced by the `MSBuild` target call below 😁

dministro (approved):

PR #34, opened 2023-02-17
OSOE-568: Remove the chalk dependency from scripts\handle-error.js

### LINKED ISSUE #30: Remove the `chalk` dependency from scripts\handle-error.js (OSOE-568)
Error handling has to be reliable. The mentioned script should be such a reliable component, so let's remove any external dependencies so it won't break.

To achieve that, extract the output coloring of warnings and errors into an optional "formatter" callback that can be passed by the caller. In the calling script, chalk should be a non-optional dependency again.

Note: The manual output coloring is only necessary when executing NE scripts in the VS Task Runner Explorer. Everywhere else it works out of the box.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-568)

adelgeiszt:
[OSOE-568](https://lombiq.atlassian.net/browse/OSOE-568)
Fixes #30

[OSOE-568]: https://lombiq.atlassian.net/browse/OSOE-568?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions/scripts/handle-error.js
+++ b/Lombiq.NodeJs.Extensions/scripts/handle-error.js
@@ -6,16 +6,15 @@ const os = require('os');
 
 function handleColoring(output, type, errorColoring, warningColoring) {
     let error = type === 'error';
-
-    if (error && errorColoring  === 'function') {
-        return errorColoring(output);
-    } 
-    else if (!error && warningColoring === 'function') {
-        return warningColoring(output);
+  
+    if (error && typeof errorColoring === 'function') {
+      return errorColoring(output);
+    } else if (!error && typeof warningColoring === 'function') {
+      return warningColoring(output);
     }
     
     return output;
-}
+  }
 
 function handleErrorObjectInner(error, type, defaultCode, errorColoring, warningColoring) {
     if (!error) {
--- a/Lombiq.NodeJs.Extensions/scripts/handle-error.js
+++ b/Lombiq.NodeJs.Extensions/scripts/handle-error.js
@@ -4,16 +4,20 @@
 
 const os = require('os');
 
-// Treat this dependency as optional because it's not available everywhere.
-let chalk;
-try {
-    chalk = require('chalk');
-}
-catch {
-    chalk = false;
+function handleColoring(output, type, errorColoring, warningColoring) {
+    let error = type === 'error';
+
+    if (error && errorColoring  === 'function') {
+        return errorColoring(output);
+    } 
+    else if (!error && warningColoring === 'function') {
+        return warningColoring(output);
+    }
+    
+    return output;
 }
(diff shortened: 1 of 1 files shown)


Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -4,16 +4,19 @@
...
+      return errorColoring(output);
+    } else if (!error && typeof warningColoring === 'function') {
+      return warningColoring(output);
+    }
+
+    return output;
 }
 
-function handleErrorObjectInner(error, type, defaultCode) {
+function handleErrorObjectInner(error, type, defaultCode, errorColoring, warningColoring) {
```
0liver, line 19: You've added two new parameters here, but don't use them anywhere.

I was just looking at the places where we would need to add the coloring callbacks, and I'm more and more convinced that this whole issue does not really make sense to pull through to the end, because it will make the code much more complicated and duplicated.

Coloring the output should be done where it currently is - when printing the output.

Please abandon this issue, close the PR, and set the value to something that more closely reflects your effort up until now. Then, resolve the issue with "Won't Do" and point to this comment.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -4,16 +4,19 @@
...
 
-// Treat this dependency as optional because it's not available everywhere.
-let chalk;
-try {
-    chalk = require('chalk');
-}
-catch {
-    chalk = false;
+function handleColoring(output, type, errorColoring, warningColoring) {
+    let error = type === 'error';
```
0liver, line 8: `let` should be `const` here.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -4,16 +4,19 @@
...
+
```
0liver, line 9: There's trailing whitespace in this line.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -4,16 +4,19 @@
...
-try {
-    chalk = require('chalk');
-}
-catch {
-    chalk = false;
+function handleColoring(output, type, errorColoring, warningColoring) {
+    let error = type === 'error';
+  
+    if (error && typeof errorColoring === 'function') {
+      return errorColoring(output);
```
0liver, line 11: Indentation should be 4 spaces, not 2. Please change this in VS, if you're using that.

0liver (changes requested):

PR #36, opened 2023-02-28
OSOE-559: Make pnpm retries more robust to prevent errors - 2nd try

0liver:
This reverts commit 9b18545c0afeeb238c5d22ef087dd689bf7d428e.

### DIFF
--- a/.github/workflows/publish-nuget.yml
+++ b/.github/workflows/publish-nuget.yml
@@ -10,3 +10,9 @@ jobs:
     uses: Lombiq/GitHub-Actions/.github/workflows/publish-nuget.yml@dev
     secrets:
       API_KEY: ${{ secrets.DEFAULT_NUGET_PUBLISH_API_KEY }}
+    with:
+      # Ignore NU5110 and NU5111 because they complain about .ps1 files outside of the tools folder. We don't want or
+      # need them there, though.
+      dotnet-pack-ignore-warning: >
+        NU5110
+        NU5111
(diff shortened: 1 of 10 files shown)


dministro:
If I understand well, the core issue is synchronization.

I would suggest a different approach instead of polling lock files in a loop because I feel scalability problems in the polling approach.
Will the number of retry enough if the number of projects is growing?

Here are some links for inspiration: 
https://stackoverflow.com/questions/62821122/msbuild-how-to-create-critical-section
https://gist.github.com/dedale/675ec80313f2a70266deb0ab78a0e2c6


Also you forget to mention powershell in docs.

dministro (changes requested):
Please mention powershell in docs.

Also check my comment: https://github.com/Lombiq/NodeJs-Extensions/pull/36#issuecomment-1453421531

0liver:
I didn't mention PowerShell, because if it's not available, I will fallback to just retrying: https://github.com/Lombiq/NodeJs-Extensions/pull/36/files#diff-3a625bfa2aa0523fa32b46b74cdd39519196c3b9744bc3dc192ac2f9d6e3ae61R85.

0liver:
> If I understand well, the core issue is synchronization.

That's correct.

Thanks for both of the links you've added! I strayed from using a separate DLL for that task, but I guess it's not all that scary.

But even using a Mutex, we would still need to specify a timeout to avoid hanging, so there's always a non-zero probability that this might fail.

I think the current solution is good enough; which doesn't mean that it cannot be improved upon. But I would like to use this version to avoid at least most of the build errors that are currently popping up.

dministro:
Ok then I added #37

0liver:
Closing in favor of https://github.com/Lombiq/NodeJs-Extensions/pull/38.

PR #38, opened 2023-03-07
OSOE-559: Serialize "pnpm link" calls to avoid concurrency errors

0liver:

### DIFF
--- a/Lombiq.NodeJs.Extensions/ExecWithMutex.cs
+++ b/Lombiq.NodeJs.Extensions/ExecWithMutex.cs
@@ -20,34 +20,31 @@ public class ExecWithMutex : Exec
     /// <inheritdoc />
     public override bool Execute()
     {
-        var timeout = TimeSpan.FromMinutes(1);
-        var stopwatch = Stopwatch.StartNew();
+        var timeout = TimeSpan.FromSeconds(60);
 
-        while (stopwatch.Elapsed < timeout)
+        Log.LogMessage(MessageImportance.Normal, "Waiting for {0}", MutexName);
+
+        using var mutex = new Mutex(initiallyOwned: false, MutexName);
+        try
         {
-            using var mutex = new Mutex(false, MutexName);
-            try
-            {
-                // Suppressing a false positive. We're releasing the lock in the _finally_ branch.
+            var stopwatch = Stopwatch.StartNew();
+
+            // Suppressing a false positive. We're releasing the lock in the _finally_ branch.
 #pragma warning disable S2222 // Locks should be released on all paths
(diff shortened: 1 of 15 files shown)


Lombiq.NodeJs.Extensions/ExecWithMutex.cs
```diff
@@ -0,0 +1,53 @@
...
+    /// <summary>
+    /// Gets or sets mutex name.
+    /// </summary>
+    [Required]
+    public string MutexName { get; set; }
+
+    /// <inheritdoc />
+    public override bool Execute()
+    {
+        var timeout = TimeSpan.FromSeconds(60);
```
dministro: Can we make this configurable?

0liver: Where would you configure it?

dministro: As a `msbuild` parameter, if I remember well `dotnet build /p:PARAM=VALUE` or something similar.

0liver: Why would you want to configure it?

dministro: Because this seems something where we can tune the build process if the 60 seconds is not enough in a bigger solution.

0liver: While I'd argue that this is overkill right now, I do agree that making it configurable helps with discoverability. So, I'll add an MSBuild property for that which can be overridden when needed.

Lombiq.NodeJs.Extensions/ExecWithMutex.cs
```diff
@@ -0,0 +1,59 @@
...
+        try
+        {
+            var stopwatch = Stopwatch.StartNew();
+
+            // Suppressing a false positive. We're releasing the lock in the _finally_ branch.
+#pragma warning disable S2222 // Locks should be released on all paths
+            if (mutex.WaitOne(timeout))
+#pragma warning restore S2222 // Locks should be released on all paths
+            {
+                Log.LogMessage(MessageImportance.Normal, "Acquired {0} after {1}", MutexName, stopwatch.Elapsed);
```
dministro, line 43: Impressing.
image

dministro (approved):

Lombiq.NodeJs.Extensions/ExecWithMutex.cs
0liver: I'm glad you checked 😁

PR #43, opened 2023-03-30
OSOE-593: Re-add additional logging after releasing v1.2.2

### LINKED ISSUE #39: MSB3073: The command "node scripts/run-pnpm-script "<ProjectName>" build" exited with code -1. (OSOE-593)
Here's an example of the error:

https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/4388557195/jobs/7685183926#step:6:114

The cause of the error seems to be a missing script in the _.bin_ folder of the Node.js Extensions NPM package installation, which in case of OSOCE is e.g. _C:\\Projects\\Lombiq\\Osoce\\node_modules\\.nx\\node_modules\\.bin_.

The specific missing script varies. In the above build, `postcss` was missing, but there have been builds where `sass` was missing instead.

### What to do

We need to investigate why `pnpm install` does not create these scripts properly at the given location. And then, we'll have to fix it.

### Ideas

- Add the `--report-summary` switch to all `pnpm run` calls (see https://github.com/pnpm/pnpm/releases/tag/v7.28.0).
- Remove the ` --prefer-offline` flag from the `pnpm install` call in _Lombiq.NodeJs.Extensions\build\NodeJsExtensionsNpmPackage.targets_
- Repeat the `pnpm install` call.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-593)

0liver:
[OSOE-593](https://lombiq.atlassian.net/browse/OSOE-593)
Fixes #39

[OSOE-593]: https://lombiq.atlassian.net/browse/OSOE-593?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


dministro (approved):

Lombiq.NodeJs.Extensions.Tests/MutexReaderWriterTest.cs
```diff
@@ -0,0 +1,83 @@
...
+        var timeout = TimeSpan.FromMilliseconds(MaxWaitTimeMs);
+        var tasks = new List<Task>();
+        for (int i = 1; i <= ReaderWriterCount; i++)
+        {
+            var action = RandomNumberGenerator.GetInt32(0, 2) < 1 ? Reader(i, timeout) : Writer(i, timeout);
+            tasks.Add(Task.Run(action));
+        }
+
+        Task.WaitAll(tasks.ToArray());
+    }
```
dministro: Please use meaningful names (`i, Reader, Writer`).
```suggestion
    public void SyncWriterAndReaders()
    {
        var timeout = TimeSpan.FromMilliseconds(MaxWaitTimeMs);
        var tasks = new List<Task>();

        for (var actionIndex = 1; actionIndex <= ReaderWriterCount; actionIndex++)
        {
            var action = RandomNumberGenerator.GetInt32(0, 2) < 1
                ? CreateReaderAction(actionIndex, timeout)
                : CreateWriterAction(actionIndex, timeout);
            tasks.Add(Task.Run(action));
        }

        Task.WaitAll(tasks.ToArray());
    }
```

Lombiq.NodeJs.Extensions.Tests/MutexReaderWriterTest.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                _testOutputHelper.WriteLine($" - Reader {i} executing");
+                Thread.Sleep(RandomNumberGenerator.GetInt32((int)(ReaderExecutionTimeMs * 0.8), (int)(ReaderExecutionTimeMs * 1.25)));
+                return true;
+            },
+            (message, mutexName) => _testOutputHelper.WriteLine($" - Reader {i} waiting: " + message, mutexName),
+            (_, _) => { });
+
+        _testOutputHelper.WriteLine($"<- Reader {i}");
+    };
```
dministro: Please use meaningful name (`i`).
You can remove the `SuppressMessage` if you are using `ITestOutputHelper.WriteLine(string format, params object[] args)` to format the string.

`mutexName` is an `object[]` type, so some name change is necessary. Eg. `args`.
Why `int * float -> int` required here? Can we use constants instead?
Where is the assertion? What is the expected result of this test? As I see this test will be a success in all cases.

Same below.

```suggestion
    private const int ReaderExecutionTimeMinMs = 800;
    private const int ReaderExecutionTimeMaxMs = 1250;
    public Action CreateReaderAction(int actionIndex, TimeSpan timeout) => () =>
    {
        _testOutputHelper.WriteLine("-> Reader {0}", actionIndex);
        // Add some random wait time to mix reader and writer threads.
        Thread.Sleep(RandomNumberGenerator.GetInt32(1000));

        new SharedMutex(MutexName, timeout)
            .Execute(
                () =>
                {
                    _testOutputHelper.WriteLine(" - Reader {0} executing", actionIndex);
                    Thread.Sleep(RandomNumberGenerator.GetInt32(ReaderExecutionTimeMinMs, ReaderExecutionTimeMaxMs));

                    return true;
                },
                (message, mutexName) =>
                    _testOutputHelper.WriteLine(" - Reader {0} waiting: {1}", i, string.Format(CultureInfo.InvariantCulture, message, mutexName)),
                (_, _) => { })
            .ShouldBeTrue();

        _testOutputHelper.WriteLine("<- Reader {0}", actionIndex);
    };
```

Lombiq.NodeJs.Extensions/CustomExecTasks/ExclusiveMutex.cs
```diff
@@ -0,0 +1,63 @@
...
+    public int RetryIntervalMs { get; set; } = 100;
+    public int WaitTimeMs { get; set; } = 1000;
+
+    public ExclusiveMutex(string mutexName, TimeSpan timeout)
+    {
+        _mutexName = mutexName;
+        _timeout = timeout;
+    }
+
+    public bool Execute(Func<bool> functionToExecute, Action<string, object[]> logWait, Action<string, object[]> logError)
```
dministro: Can we use default values for `logWait` and `logError`?

```suggestion
    public bool Execute(Func<bool> functionToExecute, Action<string, object[]> logWait = null, Action<string, object[]> logError = null)
```

Lombiq.NodeJs.Extensions/CustomExecTasks/ICustomMutex.cs
dministro, line 1: Why this `interface` is required? Not used.

Lombiq.NodeJs.Extensions/CustomExecTasks/ExecWithMutex.cs
```diff
@@ -0,0 +1,83 @@
...
+                    (message, args) => Log.LogMessage(message, args),
+                    (message, args) => Log.LogError(message, args));
+            case MutexAccess.Undefined:
+            default:
+                const string errorMessage =
+                    $"{nameof(Access)} needs to be set to {nameof(MutexAccess.Shared)} or {nameof(MutexAccess.Exclusive)}!";
+                // We use "Access" as the parameter name although it's technically a property. Still good.
+#pragma warning disable S3928 // Parameter names used into ArgumentException constructors should match an existing one.
+                throw new ArgumentOutOfRangeException(nameof(Access), errorMessage);
+#pragma warning restore S3928 // Parameter names used into ArgumentException constructors should match an existing one.
```
dministro: Please create a new exception for this state and remove the suppression.

Lombiq.NodeJs.Extensions/CustomExecTasks/SharedMutex.cs
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    public bool Execute(Func<bool> functionToExecute, Action<string, object[]> logWait, Action<string, object[]> logError)
+    {
+        var count = 1;
+        var stopwatch = Stopwatch.StartNew();
+        using var mutex = new Mutex(initiallyOwned: false, _mutexName);
+        while (stopwatch.Elapsed <= _timeout)
+        {
+            if (mutex.WaitOne(0))
```
dministro: Can we use `mutex.WaitOne(RetryIntervalMs)` instead? If so, we can remove `Thread.Sleep(RetryIntervalMs);`

dministro (changes requested):

Lombiq.NodeJs.Extensions/CustomExecTasks/SharedMutex.cs
0liver: No. The point is to release the mutex asap, but I've now realized that I don't actually do that!

Will refactor this code.

Lombiq.NodeJs.Extensions/CustomExecTasks/ExecWithMutex.cs
0liver: I'll just use an `ArgumentException` instead - that should suffice.

Lombiq.NodeJs.Extensions/CustomExecTasks/MutexAccess.cs
```diff
@@ -0,0 +1,8 @@
+namespace Lombiq.NodeJs.Extensions.CustomExecTasks;
+
+public enum MutexAccess
+{
+    Undefined = 0,
+    Exclusive = 1,
+    Shared = 2,
+}
```
dministro: I think enum values are not required. Can you agree?

```suggestion
public enum MutexAccess
{
    Undefined,
    Exclusive,
    Shared,
}
```

dministro (changes requested):
Please check my suggestion: https://github.com/Lombiq/NodeJs-Extensions/pull/43#discussion_r1166000953

Lombiq.NodeJs.Extensions/CustomExecTasks/MutexAccess.cs
0liver: I've used both versions in the past. In the current situation the values are not needed, for sure.

They should be used when storing enums as ints e.g. in a DB - then not having immutable values assigned can be a great way to shoot yourself in the foot when you add a new enum member somewhere in the middle (or the beginning 😁).

dministro: > They should be used when storing enums as ints e.g. in a DB - then not having immutable values assigned can be a great way to shoot yourself in the foot when you add a new enum member somewhere in the middle (or the beginning 😁).

Agree 😄

dministro (approved):

PR #57, opened 2023-07-14
OSOE-430: Add support for single-instance NE on Linux

### LINKED ISSUE #17: Support a single NE instance per solution on Linux, too (OSOE-430)

This should be fixed.

0liver:
Fixes #17


Piedone:
We'll open a new PR for this.

PR #58, opened 2023-07-17
OSOE-603: Upgrade to pnpm@8

### LINKED ISSUE #42: Upgrade to pnpm@8; drop support for Node.js 14 (OSOE-603)
This should be done once we've upgraded all internal projects and developers to Node.js 16, at least.

There's at least one catch, though. Citing the [v8.0.0 release notes](https://github.com/pnpm/pnpm/releases/tag/v8.0.0):

> A lockfile is generated even for projects with no dependencies.

Also, we should wait a bit until at least a few bug fixes have been incorporated.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-603)

sarahelsaig:
[OSOE-603](https://lombiq.atlassian.net/browse/OSOE-603)
Fixes #42

[OSOE-603]: https://lombiq.atlassian.net/browse/OSOE-603?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.NodeJs.Extensions/scripts/add-dev-dependencies.js
Piedone, line 1: image

Piedone (changes requested):

Piedone (approved):

Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets
```diff
@@ -4,7 +4,7 @@
   <Target Name="EnablePnpm">
     <!-- Enable the latest version of pnpm@8 using explicit version number. -->
     <!-- Ignore sporadic errors when this is run by multiple threads at the same time. -->
-    <Exec Command="corepack enable &amp;&amp; corepack prepare pnpm@8.6.8 --activate"
+    <Exec Command="corepack enable &amp;&amp; corepack prepare pnpm@8.6.9 --activate"
```
Piedone: Why not use 8.6.12, the latest version?

Lombiq.NodeJs.Extensions/pnpm-lock.yaml
```diff
@@ -1,5 +1,6639 @@
-lockfileVersion: '6.0'
+lockfileVersion: 5.4
```
Piedone: Why the downgrade form 6.0? The root one in OSOCE is 6.0.

Lombiq.NodeJs.Extensions/scripts/add-dev-dependencies.js
```diff
@@ -6,22 +6,53 @@ const { EOL } = require('os');
 
 const { handleErrorObject } = require('./handle-error');
 
+const currentDevDependencies = JSON.parse(fs.readFileSync('package.json')).devDependencies;
+function isGreaterThanCurrent([name, version]) {
```
Piedone:
```suggestion
const currentDevDependencies = JSON.parse(fs.readFileSync('package.json')).devDependencies;

function isGreaterThanCurrent([name, version]) {
```

Lombiq.NodeJs.Extensions/scripts/add-dev-dependencies.js
```diff
@@ -6,22 +6,53 @@ const { EOL } = require('os');
...
+
+    return false;
+}
+
 const configPath = path.resolve(__dirname, '..', 'package.json');
 const nxDevDependencies = JSON.parse(fs.readFileSync(configPath)).devDependencies;
 const eslintPackages = Object
     .entries(nxDevDependencies)
-    .filter(([name]) => name.startsWith('eslint'))
+    .filter(([name, version]) => name !== 'jest' && version)
```
Piedone, line 36: Why is jest excluded?

Piedone (changes requested):

Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets
sarahelsaig: 8.6.9 was the latest last month when I updated that :/

Lombiq.NodeJs.Extensions/pnpm-lock.yaml
sarahelsaig: I think it must've been a merge mistake. I've regenerated it.

Lombiq.NodeJs.Extensions/scripts/add-dev-dependencies.js
sarahelsaig: Jest is used [for testing](https://github.com/Lombiq/NodeJs-Extensions/blob/e941e63e0d96a6fb805d9971210ff7695e801275/Lombiq.NodeJs.Extensions/package.json#L83), it's not a dependency we have to inherit. This way we can add other packages too. I used this to [add semver](https://github.com/Lombiq/NodeJs-Extensions/commit/fc934f7f874c49b8d0c878918e656f1cf537b470), but in the end I found that semver [wasn't a right fit for my needs](https://github.com/Lombiq/NodeJs-Extensions/pull/58/files#diff-fc8d3f2f7941a60a5eafd9906e253ec01fb12451fd98caef357a22852b582312R14-R19) so I reverted the package but kept this change because it seems more in spirit for what we want with this script.

Piedone: Ah, clever, thanks.

Piedone (approved):

Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets
0liver: Just FYI, once you force users to use at least v16.17 of Node.js, you'll be able to use the [dist-tag](https://docs.npmjs.com/cli/v9/commands/npm-dist-tag) `latest-8` to cover all patch and minor version updates, as mentioned here: https://www.npmjs.com/package/pnpm?activeTab=versions ([docs](https://github.com/nodejs/node/pull/44646), [more docs](https://github.com/nodejs/node/pull/44646)).

Piedone: Thanks for the remark! Not having open version selectors lets us avoid surprise breakage like https://github.com/Lombiq/Orchard-Base-Theme/issues/84 though.

PR #64, opened 2023-08-11
OSOE-669: Fixing "Passing percentage units to the global abs() function is deprecated." error

### LINKED ISSUE #84: "Passing percentage units to the global abs() function is deprecated." error (OSOE-669)
During build, you'll see this error, like [here](https://github.com/Lombiq/Open-Source-Orchard-Core-Extensions/actions/runs/5801147267/job/15802216327#step:6:144). As seen from this build, the same code that correctly built two days ago fails today.

- This became an issue on its own, without us changing the code. Something in the rest of the ecosystem changed. Perhaps we have an open version in Node.js Extensions for the Sass compiler and it got updated? That should be fixed, then. 
- This looks like a Bootstrap bug: https://github.com/twbs/bootstrap/issues/39028. However, we don't want to update, see previous point. 
- I don't see it locally.

[Jira issue](https://lombiq.atlassian.net/browse/OSOE-669)

Piedone:
[OSOE-669](https://lombiq.atlassian.net/browse/OSOE-669)

Fixes https://github.com/Lombiq/Orchard-Base-Theme/issues/84.

[OSOE-669]: https://lombiq.atlassian.net/browse/OSOE-669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions/package.json
+++ b/Lombiq.NodeJs.Extensions/package.json
@@ -20,7 +20,7 @@
     "postcss": "^8.4.18",
     "postcss-cli": "^9.1.0",
     "rimraf": "^3.0.2",
-    "sass": "^1.55.0",
+    "sass": "1.59.3",
     "stylelint": "^14.14.0",
     "stylelint-config-standard-scss": "^4.0.0",
     "terser": "^5.15.1",
--- a/Lombiq.NodeJs.Extensions/pnpm-lock.yaml
+++ b/Lombiq.NodeJs.Extensions/pnpm-lock.yaml
@@ -3437,10 +3437,6 @@ packages:
     engines: {node: '>= 4'}
     dev: false
 
-  /immutable/4.1.0:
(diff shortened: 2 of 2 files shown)


Lombiq.NodeJs.Extensions/package.json
```diff
@@ -20,7 +20,7 @@
     "postcss": "^8.4.18",
     "postcss-cli": "^9.1.0",
     "rimraf": "^3.0.2",
-    "sass": "^1.55.0",
+    "sass": "1.59.3",
```
MZole, line 23: Shouldn't this be `"^1.59.3"` to allow updates within the same minor and patch versions?

MZole (changes requested):

Lombiq.NodeJs.Extensions/package.json
Piedone: That allows minor version updates too, i.e. 1.60.x too. That has bitten us here and that's specifically what I want to avoid. (I'd rather have everything as a fixed version TBH, surprise updates are not good, as we've just seen.)

MZole: I see, thank you!

PR #74, opened 2023-10-06
OSOE-430: Support a single NE instance per solution on Linux, too, in Lombiq.NodeJs.Extensions

### LINKED ISSUE #17: Support a single NE instance per solution on Linux, too (OSOE-430)

sarahelsaig:
Fixes #17

### DIFF
--- a/Lombiq.NodeJs.Extensions/scripts/copy-assets.js
+++ b/Lombiq.NodeJs.Extensions/scripts/copy-assets.js
@@ -39,7 +39,7 @@ async function copyFilesFromConfig(config) {
                         const pathPattern = path.join(directoryToCopy, pattern);
                         const targetPath = (process.platform === 'win32')
                             ? assetsGroup.target
-                            : path.normalize(path.join(projectPath, assetsGroup.target));
+                            : path.normalize(path.resolve(projectPath, assetsGroup.target));
                         const sourceAndTargetPaths = [pathPattern, targetPath];
 
                         // We want to copy all files matched by the given pattern into the target folder mirroring the
--- a/Lombiq.NodeJs.Extensions/scripts/copy-assets.js
+++ b/Lombiq.NodeJs.Extensions/scripts/copy-assets.js
@@ -28,7 +28,7 @@ async function copyFilesFromConfig(config) {
         .map((assetsGroup) => assetsGroup.sources.map((assetSource) => {
             // Normalize the relative path to the directory to remove trailing slashes and straighten out any anomalies.
             const projectPath = getProjectDirectory();
-            const directoryToCopy = path.normalize(path.join(projectPath, assetSource));
+            const directoryToCopy = path.normalize(path.resolve(projectPath, assetSource));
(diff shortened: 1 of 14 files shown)


sarahelsaig:
Note the last commit (https://github.com/Lombiq/NodeJs-Extensions/pull/74/commits/6d8e16636f70a71eefa14a28312fe527182cbea3) that contains the fixes to various ESlint warnings, including some that aren't related to this issue. (because the auxiliary scripts in _Lombiq.NodeJs.Extensions/scripts/_ that are only used by node during the compilation don't get linted automatically)

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -26,7 +27,8 @@ async function copyFilesFromConfig(config) {
     return Promise.all(config
         .map((assetsGroup) => assetsGroup.sources.map((assetSource) => {
             // Normalize the relative path to the directory to remove trailing slashes and straighten out any anomalies.
-            const directoryToCopy = path.normalize(assetSource);
+            const projectPath = getProjectDirectory();
+            const directoryToCopy = path.normalize(path.join(projectPath, assetSource));
```
dministro: `path.resolve()` seems a better choice to me instead of `path.join()` in case of absolute `assetSource`

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -35,7 +37,10 @@ async function copyFilesFromConfig(config) {
                 .then(
                     () => {
                         const pathPattern = path.join(directoryToCopy, pattern);
-                        const sourceAndTargetPaths = [pathPattern, assetsGroup.target];
+                        const targetPath = (process.platform === 'win32')
+                            ? assetsGroup.target
+                            : path.normalize(path.join(projectPath, assetsGroup.target));
```
dministro: `path.resolve()` same reason like above.

Lombiq.NodeJs.Extensions/scripts/get-path.js
```diff
@@ -46,52 +50,48 @@ function getSolutionDir(initialDirectory) {
...
     const config = getConfig({ directory: initialDirectory, verbose: verbose });
-    let effectiveDir = config[type][location];
-
-    if (!effectiveDir) return undefined;
 
-    if (effectiveDir === solutionFolderMarker) {
-        effectiveDir = getSolutionDir(initialDirectory);
-    }
+    if (!config) throw new Error(`Config ${JSON.stringify({ directory: initialDirectory, verbose: verbose })} is missing.`);
+    if (typeof config[type] !== 'object' || !config[type][location]) return null;
```
dministro: Optional chaining:
```suggestion
    if (!config?.[type]?.[location]) return null;
```

dministro (changes requested):

dministro (approved):

PR #78, opened 2023-10-31
OSOE-430: Support a single NE instance per solution on Linux, too, in Lombiq.NodeJs.Extensions (reapply #74)

### LINKED ISSUE #17: Support a single NE instance per solution on Linux, too (OSOE-430)

sarahelsaig:
Fixes #17.

Re-applying https://github.com/Lombiq/NodeJs-Extensions/pull/74 that was reverted due to some unforeseen build errors in CI builds.


Lombiq.NodeJs.Extensions/Lombiq.NodeJs.Extensions.targets
```diff
@@ -65,7 +65,8 @@
           Outputs="$(PnpmLinkStamp)">
     <!-- Store the pnpm link command in a property for easy re-use. -->
     <PropertyGroup>
-      <PnpmLinkCommand>pnpm link "$(NodeJsExtensionsNpmPackageSourcePath)"</PnpmLinkCommand>
+      <PnpmLinkCommand Condition="'$([MSBuild]::IsOSUnixLike())' == 'false'">pnpm link "$(NodeJsExtensionsNpmPackageSourcePath)"</PnpmLinkCommand>
+      <PnpmLinkCommand Condition="'$([MSBuild]::IsOSUnixLike())' == 'true'">node -e "const max = 5; for (let i = 0; i !== max; i++) { try { const proc = require('child_process').spawnSync('pnpm', [ 'link', process.argv[1] ]); proc.output.forEach(output => console.log('' + output)); if (proc.error || proc.status !== 0) throw proc; break; } catch (e) { if (i + 1 === max) process.exit(1); } }" "$(NodeJsExtensionsNpmPackageSourcePath)"</PnpmLinkCommand>
```
dministro, line 69: No better way? Just retry?

sarahelsaig: This problem appears inconsistently and only in CI. I have yet to reproduce it reliably. I suspect a timing issue and probably a bug within PNPM, but since I can't give repro instructions there is no use opening an issue about it. So far every time build failed like this, a single retry reliably solved the issue.

dministro (approved):

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -9,55 +9,68 @@ const util = require('util');
...
-                    () => handleErrorObject({
-                        code: 'NE0031',
-                        path: 'AssetCopy',
-                        message: `The directory "${directoryToCopy}" cannot be accessed to copy files from.`,
-                    }));
+                    // See https://github.com/calvinmetcalf/copyfiles#programic-api for more details.
+                    return copyfiles(sourceAndTargetPaths, { verbose: verbose, up: depth }, () => {});
+                },
+                () => handleErrorObject({
+                    code: 'NE31',
```
dministro, line 55: Is this intended? `NE0031` -> `NE31`

sarahelsaig: Yes. All other Lombiq.NodeJs.Extensions error codes are in the NExx format were xx is 2 digits, so I'm sure the original NE0031 was a mistake.

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
```diff
@@ -9,55 +9,68 @@ const util = require('util');
...
+                    path: 'AssetCopy',
+                    message: `The directory "${directoryToCopy}" cannot be accessed to copy files from.` +
+                        JSON.stringify({ pattern: pattern, assetSource: assetSource, currentDirectory: process.cwd() }),
+                }));
         }))
         .reduce((previousArray, currentArray) => [...previousArray, ...currentArray], []));
 }
 
 (async function main() {
```
dministro, line 64: This looks a bit weird, I know it's working.
`(async function main() { await ... })()`

dministro (approved):

Lombiq.NodeJs.Extensions/scripts/copy-assets.js
sarahelsaig: It's an asynchronous [immediately invoked function expression (IIFE)](https://developer.mozilla.org/en-US/docs/Glossary/IIFE).

You can't use use `await` in the top level of a node script (at least not without the `--experimental-repl-await` command line switch), but you can use it inside an async function and then it returns a Promise. Any promises you return from a top level call, node will await before exiting exiting the interpreter like this:
image

dministro: Yes, I know, thanks to the `nodejs` event scheduler we can do this. The `node` process doesn't exit while any pending event is in the queue, but it looks weird anyway.

PR #91, opened 2024-01-18
OSOE-751: Upgrade to Orchard Core 1.8

Psichorex:
[OSOE-751](https://lombiq.atlassian.net/browse/OSOE-751)

[OSOE-751]: https://lombiq.atlassian.net/browse/OSOE-751?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
+++ b/Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
@@ -4,6 +4,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>7.3</LangVersion>
     <DefaultItemExcludes>$(DefaultItemExcludes);.git*;node_modules\**</DefaultItemExcludes>
   </PropertyGroup>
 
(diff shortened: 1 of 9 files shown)


Piedone (approved):

Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
```diff
@@ -4,6 +4,7 @@
 
   <PropertyGroup>
     <TargetFramework>netstandard2.0</TargetFramework>
+    <LangVersion>7.3</LangVersion>
```
Piedone, line 8: This should rather be 8.0: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/configure-language-version#defaults. Also for the root project.

Piedone (changes requested):

Lombiq.NodeJs.Extensions.Samples/Lombiq.NodeJs.Extensions.Samples.csproj
Psichorex: But why? `netstandard2.0` is supported up until C# 7.3 according to the docs. C# 8 is from netstandard2.1 which will produce unsupported language feature warnings most probably.
https://stu.dev/csharp8-doing-unsupported-things/

Piedone: Ah, it's 2.**1**, indeed.

Piedone (approved):

PR #93, opened 2024-01-26
Update NVM SetupWindows.md

MarGraz:
Added the next step in the "Setting up Node.js with NVM for Windows" section, because the next section title "Additional configuration", could appear as not mandatory for a new Orchard user


Piedone:
Thank you!

PR #94, opened 2024-01-26
Update NVM SetupWindows.md longpaths reboot

MarGraz:
Suggest the user that a reboot is necessary for the changes to take effect after modifying the "longpaths" key


Piedone:
Thank you! I appreciate the granularity (and your many contributions in general!), but next time, feel free to bundle such small docs changes together in a single PR too.

PR #99, opened 2024-05-24
OSOE-744: Upgrade markdownlint to v.0.34.0

BenedekFarkas:
[OSOE-744](https://lombiq.atlassian.net/browse/OSOE-744)

Note: See the explanation for the comments added [here](https://github.com/Lombiq/GitHub-Actions/pull/351#issue-2303312865).

[OSOE-744]: https://lombiq.atlassian.net/browse/OSOE-744?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Lombiq.NodeJs.Extensions/config/lombiq.markdownlint.json
+++ b/Lombiq.NodeJs.Extensions/config/lombiq.markdownlint.json
@@ -1,3 +1,4 @@
+// When updating, also update the same configuration file accompanying the "markdown-lint" action in Lombiq.GitHub.Actions.
 {
   "default": true,
   "MD004": { "style": "dash" },
--- a/Lombiq.NodeJs.Extensions/package.json
+++ b/Lombiq.NodeJs.Extensions/package.json
@@ -14,7 +14,7 @@
     "eslint": "8.47.0",
     "if-env": "1.0.4",
     "glob": "10.3.10",
-    "markdownlint": "0.30.0",
+    "markdownlint": "0.34.0",
     "nodemon": "3.0.1",
     "npm-run-all": "4.1.5",
     "postcss": "8.4.31",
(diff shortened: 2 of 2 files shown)


Lombiq.NodeJs.Extensions/config/lombiq.markdownlint.json
```diff
@@ -1,3 +1,4 @@
+// When updating, also update the same configuration file accompanying the "markdown-lint" action in Lombiq.GitHub.Actions.
```
sarahelsaig, line 1: Are you sure this works? Last time I checked _.markdownlint.json_ has to be a standard JSON file (meaning it has to work when passed into `JSON.parse()` in Javascript), so comments will break the parser.

Lombiq.NodeJs.Extensions/package.json
```diff
@@ -14,7 +14,7 @@
     "eslint": "8.47.0",
     "if-env": "1.0.4",
     "glob": "10.3.10",
-    "markdownlint": "0.30.0",
+    "markdownlint": "0.34.0", // When upgrading, also upgrade the "markdown-lint" action in Lombiq.GitHub.Actions to a release that uses the same version of markdownlint.
```
sarahelsaig: Definitely don't place comments into the _package.json_ file. It gets overwritten if we ever install another package from the CLI, and right now it breaks if you try to run NPM or PNPM like this:
image

sarahelsaig (changes requested):

Lombiq.NodeJs.Extensions/package.json
BenedekFarkas: Dang it, this is what broke the OSOCE build, thanks...

Issue #116, opened 2024-12-29
Update NPM dependencies to latest (OSOE-954)


Piedone:
The project's NPM dependencies are outdated, and Renovate updated them. But, due to breaking changes, manual fixes are necessary; if you update all packages even to at most minor versions, you'll be greeted with the following warnings and errors in VS, necessitating fixes in other projects too:

![{1088AF12-D788-4E39-A827-9F5611B20D45}](https://github.com/user-attachments/assets/9f3b2318-2fcd-4d1b-9d0f-953122f2a386)

- [x] Open an issue branch from the branch of https://github.com/Lombiq/NodeJs-Extensions/pull/114 and once you're done with that, merge the branch of https://github.com/Lombiq/NodeJs-Extensions/pull/115 too.
- [x] Check all updated dependencies: Are they actually all needed?
- [x] Fix any compatibility issues, including in other OSOCE projects, like the Base Theme as seen in the above screenshot.
   - To fix any "require() of ES Module ... not supported" type error, you can use `await import()` inside an asynchronous method, [like here](https://github.com/Lombiq/Orchard-Vue.js/pull/173/files/fe76cda73328d5949408f4c513a4b55a5ac1b7f3..31b101d3ef51aeaa906d31848a11565ae1fca258#diff-156c97c4a751d6de8695f2c7acc0ac984910ebd31229f1084ba59921f30793e7R83).
   - When upgrading ESLint to the latest v9.x then configurations have to be migrated, [see here](https://eslint.org/docs/latest/use/getting-started#configuration:~:text=If%20you%20are%20coming%20from%20a%20version%20before%209%2E0%2E0%20please%20see%20the%20migration%20guide%2E).
   - When upgrading StyleLint "[The CommonJS Node.js API is deprecated.](https://stylelint.io/migration-guide/to-16)"
- [x] Should we pin the PNPM version in every project's `"packageManager"` property? It seems like a safe way to take dependency on a given PNPM version, which can also be auto-updated by Renovate. If indeed, then `corepack prepare pnpm` can be removed from _NodeJsExtensionsNpmPackage.targets_, the [`enable-corepack`](https://github.com/Lombiq/GitHub-Actions/blob/dev/.github/actions/enable-corepack/action.yml) action, and documented (and ideally automatically enforced somehow) that this property should be always added.
- [x] Remove the temporary "All packages" group from _renovate.json5_ here and "ESLint" group from Lombiq.VueJs.

[...]

PR #127, opened 2025-04-22
OCC-314: Lint script and style via workflow

sarahelsaig:
[OCC-314](https://lombiq.atlassian.net/browse/OCC-314)

[OCC-314]: https://lombiq.atlassian.net/browse/OCC-314?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/lint.yml
```diff
@@ -0,0 +1,55 @@
...
+        type: string
+        default: ''
+      style-paths:
+        description: Paths for Sass linting to the directories where the package.json files are located.
+        type: string
+        default: ''
+      style-css-paths:
+        description: Paths for CSS linting to the directories where the package.json files are located.
+        type: string
+        default: ''
```
Piedone: Why the references to packages.json files? In a lot of cases we can get rid of those, and even if not, these folders don't contain them.

Piedone: These could by default fetch all *.js and *.css files under the working directory.

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,53 @@
...
+
+        $LombiqNodeJsExtensions = $(Resolve-Path '${{ github.action_path }}/../../../Lombiq.NodeJs.Extensions').Path
+        Write-Output "LOMBIQ_NODEJS_EXTENSIONS=$LombiqNodeJsExtensions"
+        "LOMBIQ_NODEJS_EXTENSIONS=$LombiqNodeJsExtensions" >> $Env:GITHUB_ENV
+        "LOMBIQ_NODEJS_EXTENSIONS_GITHUB=true" >> $Env:GITHUB_ENV
+        "LOMBIQ_NODEJS_EXTENSIONS_GITHUB_ROOT=$PWD" >> $Env:GITHUB_ENV
+        "LOMBIQ_NODEJS_EXTENSIONS_WARN_AS_ERROR=true" >> $Env:GITHUB_ENV
+
+        corepack enable
+        corepack prepare pnpm@8.15.9+sha1.aa4bc353b5f59fe1f1df5d802ce049ddf7f3c60e --activate
```
Piedone, line 49: Add a note referencing this [here](https://github.com/Lombiq/NodeJs-Extensions/blob/628e818216d2728f91420e7944d649af7cf2eeb2/Lombiq.NodeJs.Extensions/build/NodeJsExtensionsNpmPackage.targets#L12) so we don't forget to keep them in sync (until hopefully we remove it: https://github.com/Lombiq/NodeJs-Extensions/issues/116).

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,53 @@
...
+    description: Paths for Sass linting to the directories where the package.json files are located.
+  style-css-paths:
+    required: false
+    default: .
+    description: Paths for CSS linting to the directories where the package.json files are located.
+
+runs:
+  using: composite
+  steps:
+    - name: Setup NodeJs Extensions
```
Piedone, line 34: Ah I see, so it keeps using NE, clever!

.github/workflows/lint.yml
Piedone, line 1: Document the existence of this workflow, and the ""Copy CSS/JS Assets" technique (if still necessary, see my other comment). Also document how the configuration works; what if I want to use the NE defaults but with some slight overrides? Will that work with live linting in the IDE in the editor windows (what works now, for JS in VS and JS and (S)CSS in VS Code)?

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,53 @@
...
+        pnpm link --global
+
+    - name: Execute Linting
+      shell: pwsh
+      run: |
+        $library = $Env:LOMBIQ_NODEJS_EXTENSIONS
+        $project = Get-Content $library/package.json | ConvertFrom-Json -AsHashtable
+        $packages = $project['devDependencies'].Keys | ForEach-Object { $version = $project['devDependencies'][$PsItem]; "$PsItem@$version" }
+
+        Invoke-NodeJsLinter -Packages $packages -LibraryPath $library -Type scripts -Paths '${{ inputs.script-paths }}'
```
Piedone: I don't understand why we need to read the `devDependencies` and pass it to the script. Why can't we just run the Node scripts after the `pnpm install` above?

Lombiq.NodeJs.Extensions/package.json
```diff
@@ -59,7 +59,8 @@
     "clean:assets": "node scripts/clean-assets || AREA=assets STEP=clean pnpm error",
     "--": "-------------------------------------------------- STYLES -------------------------------------------------",
     "build:styles": "SRC=$(node scripts/get-path scss source) && if-env SRC=! && echo \"Skipping styles\" && exit 0 || run-p --print-name --continue-on-error lint:styles compile:styles",
-    "lint:styles": "SRC=$(node scripts/get-path scss source) && node scripts/lint-styles \"$SRC/**/*.scss\" && pnpm prettier:auto-fix || AREA=styles STEP=linting pnpm error",
+    "lint:styles": "SRC=$(node scripts/get-path scss source) && node scripts/lint-styles \"$SRC\" \"**/*.scss\" && pnpm prettier:auto-fix || AREA=styles STEP=linting pnpm error",
+    "lint:styles-css": "SRC=$(node scripts/get-path css source) && node scripts/lint-styles \"$SRC\" \"**/*.css\" || AREA=styles STEP=linting pnpm error",
```
Piedone: Document this under _Styles.md_, and that we support CSS linting in general.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -27,12 +30,23 @@ function handleErrorObjectInner(error, type, defaultCode) {
...
+        return handleErrorObjectInner(error, 'error', defaultCode);
+    }
+
     const code = error.code || defaultCode;
-    const path = error.path || 'no-path';
     const message = (error.message?.toString() ?? JSON.stringify(error)).replace(/^error[ :]+/i, '');
     const line = 'line' in error && error.line !== undefined ? error.line : 1;
     const column = 'column' in error && error.column !== undefined ? error.column : 1;
 
+    if (process.env.LOMBIQ_NODEJS_EXTENSIONS_GITHUB?.toLowerCase() === 'true') {
```
Piedone:
```suggestion
    if (process.env.LOMBIQ_NODEJS_EXTENSIONS_GITHUB_ACTIONS?.toLowerCase() === 'true') {
```

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -46,6 +60,25 @@ function handleErrorObjectInner(error, type, defaultCode) {
     return error;
 }
 
+function handleErrorObjectForGitHub(type, code, message, path, line, column) {
+    const parameters = ['title=' + code];
+    const gitHubRoot = process.env.LOMBIQ_NODEJS_EXTENSIONS_GITHUB_ROOT;
```
Piedone:
```suggestion
    const gitHubActionsRoot = process.env.LOMBIQ_NODEJS_EXTENSIONS_GITHUB_ACTIONS_ROOT;
```

Lombiq.NodeJs.Extensions/scripts/handle-error.js
```diff
@@ -46,6 +60,25 @@ function handleErrorObjectInner(error, type, defaultCode) {
...
+
+    if (path) {
```
Piedone, line 70: Is this most recent failing run how an error looks like currently too? https://github.com/OrchardCMS/OrchardCore.Commerce/actions/runs/14615008579/job/41008923526 Because the error messages don't show the file and line number.

Piedone (changes requested):

.github/workflows/lint.yml
sarahelsaig: I don't want to do that, because that would be like calling ESLint or Stylelint from the solution directory. Then the config files in the project directories would not be respected.
Instead, setting the default to empty means that if you don't specify a path that type of step is skipped.

.github/actions/lint/action.yml
sarahelsaig: [We already do this in MSBuild](https://github.com/Lombiq/NodeJs-Extensions/blob/628e818216d2728f91420e7944d649af7cf2eeb2/Lombiq.NodeJs.Extensions/build/CreateGlobalConfigFiles.targets#L86-L88) using a script. For some reason I didn't think to just reuse that exist script, [but now I did](https://github.com/Lombiq/NodeJs-Extensions/pull/127/commits/bf283c3e3e932157a29f6608e9d71d884b3f7cc5) and it seems to work fine.

.github/workflows/lint.yml
Piedone: Now I see that these are paths for the parent directories of the given resources' folders. I don't know how this is best called (project folders?) but please explain that in the `description`s.

Also, add examples in the `description`s for clarity, and document how you add multiple values (and spaces after commas in there should also work).

Piedone: OK, let's leave it, but I think there should be a more convenient way than requiring the consumers to always pass in the paths.

.github/workflows/lint.yml
Piedone: I think we still need a project to include the config files, and that part of NE that copies the default ones into the solution root.

So, how about adding the ability to this project to only use it as a source of configuration files, without the build integration? Or we could have a `Lombiq.NodeJs.Extensions.Configuration` project where we split this out, and then `Lombiq.NodeJs.Extensions` depends on it.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
sarahelsaig: Very strange behavior, the error annotations are not created in the summary. I have added the path to the annotation message, so it's always available.

.github/workflows/lint.yml
sarahelsaig: > Document the existence of this workflow, and the ""Copy CSS/JS Assets" technique (if still necessary, see my other comment). Also document how the configuration works; what if I want to use the NE defaults but with some slight overrides?

I have altered the GHA action, to accept a JSON object where you can configure NE in the action call, without creating a package.json file. So now the MSBuild copy tasks are not needed.

> Will that work with live linting in the IDE in the editor windows (what works now, for JS in VS and JS and (S)CSS in VS Code)?

Nope. Since the project doesn't contain NE when executed locally, the './node_modules/nodejs-extensions/config/.eslintrc.lombiq-base.js' file won't exist. It's unlikely that you'd use this approach if the JS/CSS in the project was significantly complex, so I don't think this is important.

> I think we still need a project to include the config files, and that part of NE that copies the default ones into the solution root.

Both ESLint and Stylelint rely on external NPM packages (`airbnb-base` and `stylelint-config-standard-scss` respectively) so the only way to actually have working local config is to run (P)NPM installation before build. Then we are back at square one. I think, this is an interesting question, but far from trivial and well outside the scope of this issue.

Piedone: That would indeed makes things more complicated, let's leave it. BTW having a PNPM install is quite reliable with NPM Targets, what makes builds a pain is the JS/SCSS build. But you'd still need to install Node, though.

Lombiq.NodeJs.Extensions/scripts/handle-error.js
sarahelsaig: Here is a newer example of the lint errors: https://github.com/Lombiq/Orchard-Privacy/actions/runs/14820609336/job/41607027094

Piedone: Looks good!

.github/actions/lint/Invoke-NodeJsLinter.ps1
```diff
@@ -46,6 +50,28 @@ function Install-NodeJsPackage($LibraryPath)
...
+
+    if ($Paths.Trim().StartsWith('{'))
+    {
+        $pathItems = $Paths | ConvertFrom-Json -AsHashtable
+        $pathItems.Keys | ForEach-Object {
+            $projectPath = (Get-Item $PSItem).FullName
+            $sourcePath = Join-Path $LibraryPath 'config' 'consumer' 'package.project.json'
+            $targetPath = Join-Path $projectPath 'package.json'
+
+            # In this case we assume that there is no project.json file, otherwise the config would already be in it.
```
Piedone, line 62: Let's test for that, though, with a hard failure.

Readme.md
```diff
@@ -129,6 +129,34 @@ The `compile` script is a wrapper for the `compile:styles`, `compile:scripts` an
...
+        {
+          "src/Modules/OrchardCore.Commerce": { "scripts": { "source": "wwwroot/js" } },
+          "src/Modules/OrchardCore.Commerce.ContentFields": { "scripts": { "source": "wwwroot/js" } },
+          "src/Modules/OrchardCore.Commerce.Payment.Stripe": { "scripts": { "source": "wwwroot/js" } }
+        }'
+      styles-css: '
+        {
+          "src/Modules/OrchardCore.Commerce": { "styles": { "source": "wwwroot/css" } },
+          "src/Modules/OrchardCore.Commerce.Payment": { "styles": { "source": "wwwroot/css" } }
+        }'
```
Piedone, line 155: Also add an example of the alternate, simple comma-separated config.

.github/actions/lint/action.yml
```diff
@@ -0,0 +1,77 @@
...
+inputs:
+  scripts:
+    required: false
+    default: ''
+    description: |
+      Information about the projects where JavaScript linting should occur. Can be in two formats:
+      - A comma separated list of relative paths to the project directories. These are the paths where a package.json
+        file could be, if desired. If there is no package.json on the provided path, the default one from this
+        repository is copied and the default asset subpath (./Assets/Scripts) is assumed. Using this approach, you have
+        to copy the assets to their final path (in ./wwwroot) using a different method, for example using the Copy
```
Piedone, line 13: Add an example of the Copy task (just a copy of what you previously added) to the Readme with the simple config sample.

Piedone (changes requested):

.github/actions/lint/Invoke-NodeJsLinter.ps1
sarahelsaig: Why is that necessary?
Also with hard failure, if you lint different things (e.g. JS and CSS) for the same project, then the generated package.json from the first operation would trip the second operation. I would have to create additional cleanup logic which otherwise would not be needed.

.github/actions/lint/action.yml
sarahelsaig: Educating the reader on basic MSBuild features? Feels like mission creep.

Piedone: We tell the user here what to do, but leave figuring out the details as an exercise to the reader. Not really an exercise that's worth doing every single time.

.github/actions/lint/Invoke-NodeJsLinter.ps1
Piedone: OK, leave it like that.

Lombiq.NodeJs.Extensions/config/lombiq.markdownlint.json
```diff
@@ -2,6 +2,7 @@
   "default": true,
   "MD004": { "style": "dash" },
   "MD013": false,
+  "MD028": false,
```
sarahelsaig, line 5: Reason for disabling is that it triggers false positive when you have two unrelated, consecutive block quotes next to each other. Which is perfectly normal in GitHub [with their info box system](https://github.com/orgs/community/discussions/16925).
More info in [this issue](https://github.com/DavidAnson/markdownlint/issues/263), which was basically closed because they won't do anything about it.

Readme.md
```diff
@@ -155,6 +157,30 @@ Create a new workflow or add the following step to an existing one that's trigge
 
 You have to provide a JSON object for the `scripts` and `styles-css` inputs, where the property names are the relative paths of the projects you want to inspect, and the values become the`nodejsExtensions` properties in the temporarily generated _package.json_ files used for the linting operation. For more information, check out the workflow inputs [here](.github/workflows/lint.yml).
 
+### Simplified configuration
+
+If the files to be linted are located in the _./Assets/Scripts_ and _./Assets/Styles_ subdirectories, then you can use the simplified comma-separated format:
```
Piedone: Actually why don't have use _wwwroot/js_ and _wwwroot/css_ as defaults? It seems to me that the most common scenario will be this, but now we require the most configuration (of which most is copy-pasted) for this scenario. Or alternatively, detect these folders when the simple config is used: e.g. with the `scripts: src/Modules/OrchardCore.Commerce` config, check if there's a _src/Modules/OrchardCore.Commerce/Assets/Scripts_ folder, and if yes, lint that. If not, then check for a _src/Modules/OrchardCore.Commerce/wwwroot/js folder.

Readme.md
```diff
@@ -155,6 +157,30 @@ Create a new workflow or add the following step to an existing one that's trigge
...
+
+```yml
+  lint:
+    name: Lint Scripts and Styles
+    uses: Lombiq/NodeJs-Extensions/.github/workflows/lint.yml@dev
+    with:
+      scripts: src/Modules/OrchardCore.Commerce,src/Modules/OrchardCore.Commerce.ContentFields,src/Modules/OrchardCore.Commerce.Payment.Stripe
+      styles-css: src/Modules/OrchardCore.Commerce,src/Modules/OrchardCore.Commerce.Payment
```
Piedone: Do spaces between commas work too? They'd need to, and then change this to reflect that, since that's the standard English spelling.

Piedone (changes requested):

Readme.md
sarahelsaig: Yes, it works because the code does `$Paths.Split(',') | ForEach-Object { $PSItem.Trim() }` [here](https://github.com/Lombiq/NodeJs-Extensions/blob/6a588f03ca0baf695ba94d5cf7e479758948585c/.github/actions/lint/Invoke-NodeJsLinter.ps1#L75-L76). I've updated the readme.

Piedone (approved):

PR #144, opened 2025-10-10
OSOE-954: Update NPM dependencies to latest in Lombiq.NodeJs.Extensions

### LINKED ISSUE #116: Update NPM dependencies to latest (OSOE-954)

sarahelsaig:
[OSOE-954](https://lombiq.atlassian.net/browse/OSOE-954)
Fixes #116

[OSOE-954]: https://lombiq.atlassian.net/browse/OSOE-954?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Lombiq.NodeJs.Extensions/package.json
Piedone, line 1: Publish an alpha to NuGet and update the NuGetTest solution as well, please, and the NE references in projects that directly depend on it.

Lombiq.NodeJs.Extensions/package.json
```diff
@@ -63,7 +62,7 @@
     "lint:styles-css": "SRC=$(node scripts/get-path css target) && node scripts/lint-styles \"$SRC/**/*.css\" || AREA=styles STEP=linting pnpm error",
     "compile:styles": "SRC=$(node scripts/get-path scss source) && if-env SRC=! && echo \"Skipping styles\" && exit 0 || DEST=$(node scripts/get-path scss target) pnpm _compile:styles",
     "_compile:styles": "run-s --print-name _sass:styles _autoprefix:styles _minify:styles _rm:temp:styles || AREA=styles STEP=compilation pnpm error",
-    "_sass:styles": "sass \"$SRC:$DEST/_temp_\" --embed-sources",
+    "_sass:styles": "sass \"$SRC:$DEST/_temp_\" --embed-sources --silence-deprecation=color-functions --silence-deprecation=global-builtin --silence-deprecation=import",
```
Piedone, line 67: Why are these necessary? Only for https://github.com/Lombiq/Orchard-Base-Theme/pull/152/files#diff-72c9dda3f0866286081e38c35f9ca0bd30310190b5f508738298b8dcee5714a7?

Lombiq.NodeJs.Extensions/package.json
Piedone, line 1: Is everything on latest now?

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
Piedone, line 1: There are some ESLint warnings here:

Image

Similarly in the project config too.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
```diff
@@ -1,11 +1,13 @@
-module.exports = {
-    // Setting root=true prevents ESLint from taking into account .eslintrc files higher up in the directory tree.
-    root: true,
+const { FlatCompat } = require('@eslint/eslintrc');
```
Piedone, line 2: This is for backwards compatibility, right? But we eventually need to bite the bullet, and it's a better time now than later.

Lombiq.NodeJs.Extensions/package.json
Piedone, line 1: Shouldn't this include `"packageManager": "pnpm@8.15.9",` too?

Piedone (changes requested):

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
sarahelsaig: Not necessarily. Currently the `eslint-config-airbnb-base` package (which we use as the basis for JS linting rules) does not support the flat configuration format. Maybe they will make the change and then we won't have to bite any bullet. If we want to get rid of `FlatCompat` now, we'd have to switch over to a different eslint-config package and then decide how to handle the difference.

We can do that, but I think it should be handled in a separate issue. It may involve some in-house discussion about which package to choose and what adjustments to make.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
sarahelsaig: These only show up in VS and I don't have Windows, so I can't actually see them for myself. {lease forward me the full list of warnings you see.

EDIT: if all of the errors are from the Lombiq.NodeJs.Extensions/config directory, then I've fixed those. I don't know why we need to lint config files, but now I've added an additional workflow to lint that directory as well (previously we only linted the scripts directory in NE) so it won't be a problem in the future.

Lombiq.NodeJs.Extensions/package.json
sarahelsaig: yes

Lombiq.NodeJs.Extensions/package.json
sarahelsaig: Yes. Added comment.

Lombiq.NodeJs.Extensions/package.json
Piedone: OK, great!

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
Piedone: There were all the warnings, so now they're fixed, thanks.

We don't really need to lint these, but if the editor shows warnings then we should fix them.

Lombiq.NodeJs.Extensions/config/consumer/.eslintrc.global.js
Piedone: I see, OK then. There's https://github.com/airbnb/javascript/issues/2961 for this BTW.

Lombiq.NodeJs.Extensions/config/consumer/package.project.json
```diff
@@ -1,5 +1,6 @@
 {
   "private": true,
+  "packageManager": "pnpm@8.15.9",
```
Piedone, line 3: After a clean build, the EInvoiceValidator, UIKit, Walkthroughs projects have diffs like this:

Image

Piedone (changes requested):

